PR #6, opened 2014-10-24
Feature/fix invoice line hook + pt_br Translate

mileo:


pedrobaeza:
Patch looks good, but I cannot say about translation.

mileo:
This is the same translation of https://github.com/OCA/commission/pull/5

Sorry about the translation, we have a problem with a mirror this week and lost some branchs. 

@renatonlima Vc poderia dar uma revisada? 

Regards

sale_commission/i18n/pt_BR.po
```diff
@@ -80,7 +79,7 @@ msgstr "Resumo da fatura"
 #. module: sale_commission
 #: sql_constraint:sale.order:0
 msgid "Order Reference must be unique !"
-msgstr "Referência à ordem deve ser único!"
+msgstr "Referência à ordem deve ser única!"
```
danimaribeiro, line 82: "Referência do pedido deve ser única!" 
This one looks better.

danimaribeiro:
I added one note to one translation.
The rest looks good, feel free to merge.

sale_commission/stock_picking.py
```diff
@@ -48,7 +48,7 @@ def _invoice_line_hook(self, cr, uid, move_line, invoice_line_id, context=None):
         if context is None:
             context = {}
         agent_pool = self.pool.get('invoice.line.agent')
-        super(stock_picking, self)._invoice_line_hook(cr, uid, move_line, invoice_line_id, context=context)
+        super(stock_picking, self)._invoice_line_hook(cr, uid, move_line, invoice_line_id)
```
pedrobaeza, line 51: You shouldn't change this

foutoucour, line 51: Hey,

_invoice_line_hook doesn't take context as argument anymore in fact. Without this patch, the tests of commission fails

foutoucour:
:+1: 
The fails come from pep8 issues from the previous version.

sale_commission/stock_picking.py
pedrobaeza, line 51: OK, thanks for explaining

yvaucher:
:+1:

PR #12, opened 2015-02-12
[IMP] Changed the select of invoice adding a function that allows change...

mileo:
... the states and query conditions


pedrobaeza:
As Travis is red and no fix, I close. Feel free to reopen in case there's interest.

PR #17, opened 2015-05-15
[FIX] commission_free field not visible on view

dcorio:
the commission_free field wasn't visibile on the product form


pedrobaeza:
Hi, Davide, please check my PR (https://github.com/OCA/commission/pull/14) that contains a full refactorization of the modules, so this patch is meaningless.

dcorio:
@pedrobaeza whooopss... i didn't notice you PR. thank you so much!

PR #23, opened 2015-06-18
[8.0] sale_commission_formula

dcorio:
module developed to satisfy a customer's specific need.
it should be generic enough for the inclusion in this OCA's project.


dcorio:
I'd like to have your opinion and suggestions about this module.
thanks in advance.

nhomar:
@dcorio sorry closed by mistake :-(

sale_commission_formula/models/account_invoice.py
```diff
@@ -0,0 +1,40 @@
...
+    @api.one
+    @api.depends('commission.commission_type', 'invoice_line.price_subtotal')
+    def _get_amount(self):
+        if self.commission.commission_type == 'formula' and (
+            not self.invoice_line.product_id.commission_free and
+                self.commission):
+            self.amount = 0.0
+            formula = self.commission.formula
+            results = {'line': self.invoice_line}
+            exec(formula) in results
```
pedrobaeza: Maybe you should use safe_eval or similar to avoid the injection of malicious code

sale_commission_formula/models/sale_commission.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Commission(models.Model):
+    _inherit = 'sale.commission'
+
+    commission_type = fields.Selection(selection_add=[("formula", "Formula")])
+
+    formula = fields.Text(
+        'Formula',
+        default="""# Use 'line' to access all the invoice line attributes\n
+result = 0 # use 'result' to return the commission amount""")
```
pedrobaeza: Don't use triple quota, but simple ones enclosed on each line

sale_commission_formula/models/account_invoice.py
dcorio: Thanks Pedro, i'll change my code accordingly. I assumed exec was secure enough because i saw Odoo using it elsewhere.

pedrobaeza:
Hi, @dcorio,

I have checked your module and it works well, but there are a few issues here to solve:
- You are using the same formula for sale orders and for invoices, but passing line as a different object (account.invoice.line or sale.order.line), so there's the possibility to write a formula that works in one place but not in the other. For solving this, we can:
  - Define 2 formulas, one for the sale order and one for the account invoice.
  - Document in the same field how can we compare in the formula to see if we are at sale order or at account invoice (`if line._name == 'sale.order.line': ...`).
- The data to compute the commission can be only in the sale order, so we need a way to indicate that the amount mustn't be recomputed on account invoices.

rafaelbn:
Hi @dcorio as @pedrobaeza says we must fix the formula's problem as commissions could be in sales.order or more often in invoices. Could tell us something about this? I will test it. Thanks

dcorio:
@rafaelbn @pedrobaeza thanks for the feedback and suggestions. i'll get back to you in the very few next days.

dcorio:
@rafaelbn @pedrobaeza i finally found the time to fix those things on this module. i didn't split the formula field in two since i wouldn't make sense considering the fact that one should not apply different commission rates on SO and invoice. So i decided to better explain how to configure the commission. Right to the field you now have an explanation and a table with the most used fields for the two modules. The default formula already provides the correct structure.

I also converted the exec() call into a safe_eval one.

pedrobaeza:
Thanks for the changes. You should differentiate the icon from the generic one. You can edit the SVG with Inkscape and add a text with some formula scheme like "x = y + z" to represent the goal of this module. Tests are also welcome.

dcorio:
thank you very much @pedrobaeza for the review. could you please have a look at the code now?

dcorio:
PS: working on the tests

pedrobaeza:
Thanks, :+1: waiting for the tests

tafaRU:
@dcorio, thanks for this module!

Also please adapt the README file to https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst and update the version to https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

hurrinico:
Hi @dcorio ,
Thanks for the module
Could i ask you to rebase your branch?
Thank You in advance

ivantodorovich:
:+1: waiting for this merge!

pedrobaeza:
@dcorio, please update latest comments about README file and I'll merge this one.

dcorio:
@pedrobaeza i'll do the update this evening. sorry for the delay but i'm quite busy these days.

pedrobaeza:
It seems the rebase have failed. @dcorio, can you please check it and also squash a bit your commits to a few?

dcorio:
@pedrobaeza noticed that thanks. going to fix it

hurrinico:
@dcorio @pedrobaeza to fix this problem i have done: https://github.com/hurrinico/commission/tree/add_sale_commission_formula if you want i can do another PR in order to solve the problem

dcorio:
@hurrinico sure! this will save time

hurrinico:
@dcorio @pedrobaeza I've created this https://github.com/OCA/commission/pull/48

eLBati:
@hurrinico @dcorio can this be closed?

dcorio:
@eLBati yes, we can continue on https://github.com/OCA/commission/pull/48

PR #27, opened 2015-08-17
Handle more complex commission structures

veloutin:
- Support multiple commissions per agent
- Support commissions based on product margins
- Handle period-based commissions

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -22,7 +22,7 @@
 ##############################################################################
 {
     'name': 'Sales commissions',
-    'version': '2.0',
+    'version': '2.1',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
@@ -45,6 +45,7 @@
         "Giorgio Borelli <giorgio.borelli@abstract.it>",
         "Daniel Campos <danielcampos@avanzosc.es>",
         "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
+        "Vincent Vinet <vincent.vinet@savoirfairelinux.com>",
     ],
     "data": [
         "security/ir.model.access.csv",
@@ -56,7 +57,6 @@
         "views/settlement_view.xml",
         "wizard/wizard_settle.xml",
         "wizard/wizard_invoice.xml",
-        # "report/cc_commission_report.xml"
     ],
     "demo": [
         # 'demo/sale_agent_demo.xml',
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -28,11 +28,11 @@ class AccountInvoice(models.Model):
     _inherit = "account.invoice"
 
     @api.one
-    @api.depends('invoice_line.agents.amount')
+    @api.depends('invoice_line.commissions.amount')
     def _get_commission_total(self):
         self.commission_total = 0.0
         for line in self.invoice_line:
-            self.commission_total += sum(x.amount for x in line.agents)
+            self.commission_total += sum(x.amount for x in line.commissions)
         # Consider also purchase refunds, although not in the initial scope
         if self.type in ('out_refund', 'in_refund'):
             self.commission_total = -self.commission_total
@@ -70,34 +70,36 @@ class AccountInvoice(models.Model):
 
 
 class AccountInvoiceLine(models.Model):
+    """Invoice Line inherit to add commissions"""
     _inherit = "account.invoice.line"
 
     @api.model
-    def _default_agents(self):
-        agents = []
-        if self.env.context.get('partner_id'):
-            partner = self.env['res.partner'].browse(
-                self.env.context['partner_id'])
-            for agent in partner.agents:
-                agents.append({'agent': agent.id,
-                               'commission': agent.commission.id})
-        return [(0, 0, x) for x in agents]
+    def _default_commissions(self):
+        res = self.env['sale.commission'].get_default_commissions()
+        return [(0, 0, x) for x in res]
 
-    agents = fields.One2many(
-        comodel_name="account.invoice.line.agent",
+    commissions = fields.One2many(
+        comodel_name="account.invoice.line.commission",
         inverse_name="invoice_line", string="Agents & commissions",
         help="Agents/Commissions related to the invoice line.",
-        default=_default_agents, copy=True)
+        default=_default_commissions, copy=True)
     commission_free = fields.Boolean(
         string="Comm. free", related="product_id.commission_free",
         store=True, readonly=True)
 
 
-class AccountInvoiceLineAgent(models.Model):
-    _name = "account.invoice.line.agent"
+class AccountInvoiceLineAgentCommission(models.Model):
+    _name = "account.invoice.line.commission"
+    _description = "Commissions to be paid on an invoice line"
 
     invoice_line = fields.Many2one(
         comodel_name="account.invoice.line", required=True, ondelete="cascade")
+    commission = fields.Many2one(
+        comodel_name="sale.commission", required=True, ondelete="restrict")
+    agent = fields.Many2one(
+        comodel_name="res.partner", required=True, ondelete="restrict",
+        domain="[('agent', '=', True)]")
+
     invoice = fields.Many2one(
         comodel_name="account.invoice", string="Invoice",
         related="invoice_line.invoice_id", store=True)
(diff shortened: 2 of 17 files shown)


veloutin:
@pedrobaeza You want to take a look?

sale_commission/__openerp__.py
```diff
@@ -22,7 +22,7 @@
 ##############################################################################
 {
     'name': 'Sales commissions',
-    'version': '2.0',
+    'version': '2.1',
```
pedrobaeza, line 25: Please use new version numbering scheme: 8.0.2.1.0 (https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers)

sale_commission/models/account_invoice.py
```diff
@@ -70,34 +70,36 @@ def _refund_cleanup_lines(self, lines):
...
+        default=_default_commissions, copy=True)
     commission_free = fields.Boolean(
         string="Comm. free", related="product_id.commission_free",
         store=True, readonly=True)
 
 
-class AccountInvoiceLineAgent(models.Model):
-    _name = "account.invoice.line.agent"
+class AccountInvoiceLineAgentCommission(models.Model):
+    _name = "account.invoice.line.commission"
```
pedrobaeza, line 115: This model renaming forces to create migration scripts, so maybe it's better not to rename it.

sale_commission/models/account_invoice.py
```diff
@@ -70,34 +70,36 @@ def _refund_cleanup_lines(self, lines):
...
-                               'commission': agent.commission.id})
-        return [(0, 0, x) for x in agents]
-
-    agents = fields.One2many(
-        comodel_name="account.invoice.line.agent",
+    def _default_commissions(self):
+        res = self.env['sale.commission'].get_default_commissions()
+        return [(0, 0, x) for x in res]
+
+    commissions = fields.One2many(
```
pedrobaeza, line 104: This field renaming forces to make migration efforts. In this case, it can be only to add `oldname='agents'` attribute.

sale_commission/models/account_invoice.py
```diff
@@ -106,47 +108,34 @@ class AccountInvoiceLineAgent(models.Model):
...
     product = fields.Many2one(
         comodel_name='product.product', related="invoice_line.product_id")
-    agent = fields.Many2one(
-        comodel_name="res.partner", required=True, ondelete="restrict",
-        domain="[('agent', '=', True)]")
-    commission = fields.Many2one(
-        comodel_name="sale.commission", required=True, ondelete="restrict")
     amount = fields.Float(
         string="Amount settled", compute="_get_amount", store=True)
-    agent_line = fields.Many2many(
```
pedrobaeza, line 116: Why removing the settlement line and get only the less focus settlement object?

sale_commission/models/res_partner.py
```diff
@@ -38,16 +40,18 @@ class ResPartner(models.Model):
     agent_type = fields.Selection(
         selection=[("agent", "External agent")], string="Type", required=True,
         default="agent")
-    commission = fields.Many2one(
+    commissions = fields.Many2many(
```
pedrobaeza, line 43: Again, a migration script that moves the data present on the many2one field to the many2many one is needed.

sale_commission/models/res_partner.py
```diff
@@ -22,6 +22,8 @@
 ##############################################################################
 from openerp import models, fields, api
 
+from . import sale_commission as sc
```
pedrobaeza, line 25: Don't rename `sale_commission` to avoid confusions on code mentioning

sale_commission/models/res_partner.py
```diff
@@ -38,16 +40,18 @@ class ResPartner(models.Model):
...
+        column1="partner_id", column2="commission_id",
         help="This is the default commission used in the sales where this "
              "agent is assigned. It can be changed on each operation if "
              "needed.")
     settlement = fields.Selection(
-        selection=[("monthly", "Monthly"),
-                   ("quaterly", "Quarterly"),
-                   ("semi", "Semi-annual"),
-                   ("annual", "Annual")],
+        selection=[(sc.PERIOD_MONTH, "Monthly"),
```
pedrobaeza, line 51: This is better to directly declare the full list with the tuples as a variable, like made in other places along Odoo.

sale_commission/models/sale_commission.py
```diff
@@ -36,6 +42,62 @@ class SaleCommission(models.Model):
...
+    )
+    scope = fields.Selection(
+        selection=[("own_sales", "On his customer sales"),
+                   ("company_sales", "On company sales"),
+                   ("own_margin", "On his customers margin"),
+                   ("company_margin", "On company margin")],
+        required=True, default="own_sales",
+    )
+
+    def has_company_scope(self):
```
pedrobaeza: Why don't make this computed fields accessible with dot notation?

sale_commission/models/sale_commission.py
```diff
@@ -36,6 +42,62 @@ class SaleCommission(models.Model):
...
+        return self.scope in ("company_sales", "company_margin")
+
+    def has_own_scope(self):
+        return self.scope in ("own_sales", "own_margin")
+
+    agent_ids = fields.Many2many(
```
pedrobaeza, line 63: Don't split fields definition. It makes it harder to read.

sale_commission/models/sale_commission.py
```diff
@@ -36,6 +42,62 @@ class SaleCommission(models.Model):
...
+        string="Agents", comodel_name="res.partner",
+        relation="agent_commission_rel",
+        column2="partner_id", column1="commission_id",
+    )
+
+    @api.multi
+    def compute_sale_commission(self, sale_line):
```
pedrobaeza, line 76: This method doesn't make sense here, as it's refer to sales order line. Having computed fields very easy to handle on new API, I see no reason to use these helper methods.

sale_commission/models/settlement.py
```diff
@@ -154,25 +155,38 @@ def make_invoices(self, journal, product, date=False):
...
+                                   readonly=True)
+    invoice = fields.Many2one(comodel_name='account.invoice',
+                              string="Invoice", readonly=True)
+    agent = fields.Many2one(comodel_name="res.partner", readonly=True)
+    settled_amount = fields.Float(readonly=True)
+    commission = fields.Many2one(comodel_name="sale.commission",
+                                 readonly=True)
+
+    def init(self, cr):
+        tools.drop_view_if_exists(cr, self._table)
```
pedrobaeza, line 189: Why making this a view? I see no special reason and hits the performance. Related fields are enough for such things.

pedrobaeza:
Hi, @veloutin, thanks for the effort of making such changes. I have started reviewing the PR, but I find a lot of difficulties to get the review done because:
- There are a lot of unneeded renamings. Maybe some field's names are not the best (but they're appropiated anyways), but handling these renamings in migrations scripts is a complex thing, so we shouldn't take care of this in a middle of a version.
- This PR covers a lot of functionality (scope, multi-agents per partner, etc), and it's very difficult to review. I recommend you to add each of the new features one by one in a PR, so people can review them easily and helps each PR to be merge. If not, I'm afraid this PR will be held here for too many months.
- I don't agree with some coding decisions like replacing stored tables with some views that can hit performance.

So I ask you to rework on the changes and start with these smaller PRs.

Thanks.

sale_commission/models/settlement.py
veloutin, line 189: I made this a view because it didn't do anything more than act as a M2M between commissions and settlements and made code more complicated.

I haven't put much data in, so it may not be representative, but in my case it was still very fast. The view is quite simple and not so costly with a hash join.

pedrobaeza, line 189: But you're introducing SQL, so I prefer to avoid it. Related fields are there to give in the worst case (not stored), the same performance, but with a straight syntax and with more flexibility (not all packed in an SQL sentence).

veloutin:
With the latest changes, the use case seems quite different from the original module:
1) Commissions can be set up by periods, on sales and margins
2) Each agent can have many different commissions
3) Commissions are set automatically and not manually
4) Commission lines can belong to a single settlement
5) When cancelling an invoice, settled lines are reverted.

I am wondering, is there any way to reconcile the original behavior with this one, or would this be better suited as an alternate module?

pedrobaeza:
It can be reconciled without problem with original behaviour. We just need to provide a migration path, in combination with good default values and something that can be reviewable, so please split the changes and make smaller PRs.

max3903:
@pedrobaeza What would be the scope of each PR ?

pedrobaeza:
A good scope can be each of the points that @veloutin has pointed.

max3903:
@pedrobaeza The points listed by @veloutin are the steps of the process.

pedrobaeza:
Each point can be easily a feature to be developped isolated. I insist to easy the review.

veloutin:
@pedrobaeza I will try to split it as such and also try to avoid making several PRs with just as big changes, as it wouldn't help the review effort.

pedrobaeza:
Thank you very much. I'll review as soon you make the PRs.

max3903:
superseeded by #38

PR #28, opened 2015-08-27
[IMP] Add commission period and type

veloutin:
This is the first part of https://github.com/OCA/commission/pull/27

This PR adds support for different commission periods (month, quarter, semi, annual) and types (own sales, company sales, margin on own/company sales).

A notable change with scopes is that agents with a "company sales/margin" scoped commission will now appear in defaults for all of the company sales/invoices.

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -128,13 +128,13 @@ class AccountInvoiceLineAgent(models.Model):
     @api.depends('commission.commission_type', 'invoice_line.price_subtotal')
     def _get_amount(self):
         self.amount = 0.0
-        if (not self.invoice_line.product_id.commission_free and
-                self.commission):
-            subtotal = self.invoice_line.price_subtotal
-            if self.commission.commission_type == 'fixed':
-                self.amount = subtotal * (self.commission.fix_qty / 100.0)
-            else:
-                self.amount = self.commission.calculate_section(subtotal)
+        sign = {
+            'out_invoice': 1, 'in_invoice': -1,
+            'out_refund': -1, 'in_refund': 1,
+        }[self.invoice.type or 'out_invoice']
+        amount = self.commission.compute_invoice_commission(
+            self.invoice_line)
+        self.amount = sign * amount
 
     @api.one
     @api.depends('agent_line', 'agent_line.settlement.state', 'invoice',
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -70,24 +70,19 @@ class AccountInvoice(models.Model):
 
 
 class AccountInvoiceLine(models.Model):
+    """Invoice Line inherit to add commissions"""
     _inherit = "account.invoice.line"
 
     @api.model
-    def _default_agents(self):
-        agents = []
-        if self.env.context.get('partner_id'):
-            partner = self.env['res.partner'].browse(
-                self.env.context['partner_id'])
-            for agent in partner.agents:
-                agents.append({'agent': agent.id,
-                               'commission': agent.commission.id})
-        return [(0, 0, x) for x in agents]
+    def _default_commissions(self):
+        res = self.env['sale.commission'].get_default_commissions()
+        return [(0, 0, x) for x in res]
 
     agents = fields.One2many(
         comodel_name="account.invoice.line.agent",
         inverse_name="invoice_line", string="Agents & commissions",
         help="Agents/Commissions related to the invoice line.",
-        default=_default_agents, copy=True)
+        default=_default_commissions, copy=False)
     commission_free = fields.Boolean(
         string="Comm. free", related="product_id.commission_free",
         store=True, readonly=True)
--- a/sale_commission/models/res_partner.py
+++ b/sale_commission/models/res_partner.py
@@ -22,6 +22,13 @@
 ##############################################################################
 from openerp import models, fields, api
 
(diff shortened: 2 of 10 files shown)


sale_commission/models/account_invoice.py
```diff
@@ -128,13 +123,13 @@ def onchange_agent(self):
     @api.depends('commission.commission_type', 'invoice_line.price_subtotal')
```
pedrobaeza, line 123: Include here `invoice.type` as one of the depends

sale_commission/models/account_invoice.py
```diff
@@ -128,13 +123,13 @@ def onchange_agent(self):
...
-            subtotal = self.invoice_line.price_subtotal
-            if self.commission.commission_type == 'fixed':
-                self.amount = subtotal * (self.commission.fix_qty / 100.0)
-            else:
-                self.amount = self.commission.calculate_section(subtotal)
+        sign = {
+            'out_invoice': 1, 'in_invoice': -1,
+            'out_refund': -1, 'in_refund': 1,
+        }[self.invoice.type or 'out_invoice']
+        amount = self.commission.compute_invoice_commission(
```
pedrobaeza, line 130: Making this, you have obscured what are the depending fields for the computation, so I recommend you not to do it (that's why I didn't do it). Imagine you include another field for the computation. Having decoupled the computation method, you won't remember to add that field to the depends section. And also you don't win anything, because this computation is invoice line specific.

What you can do is to refactor a compute commission method that gets amount as parameter (or any other you need in subsequent PRs), and return the corresponding commission

sale_commission/models/account_invoice.py
pedrobaeza, line 123: Now that I'm revisiting the code, I think you should remove `commission.commission_type` from the depends section, because if you change the type in the future, old past invoices will be recalculated.

sale_commission/models/res_partner.py
```diff
@@ -22,6 +22,13 @@
 ##############################################################################
 from openerp import models, fields, api
 
+from .sale_commission import (
```
pedrobaeza, line 25: Instead of this, it's easier to have directly a variable with the list of tuples:

```
settlement_periods = [
    ("monthly", "Monthly"),
    ("quaterly", "Quarterly"),
    ("semi", "Semi-annual"),
    ("annual", "Annual")]
```

There's no problem with the translations, as they are parsed the same way.

sale_commission/models/sale_commission.py
```diff
@@ -23,6 +23,12 @@
 from openerp import models, fields, api, exceptions, _
 
 
+PERIOD_MONTH = "monthly"
```
pedrobaeza, line 26: Same comment as above

sale_commission/models/sale_commission.py
```diff
@@ -36,6 +42,70 @@ class SaleCommission(models.Model):
...
         comodel_name="sale.commission.section", inverse_name="commission")
     active = fields.Boolean(default=True)
+    period = fields.Selection(
+        selection=[(PERIOD_MONTH, "Monthly"),
+                   (PERIOD_QUARTER, "Every quarter"),
+                   (PERIOD_SEMI, "Every semester"),
+                   (PERIOD_YEAR, "Every year")],
+        required=True, default=PERIOD_MONTH,
+    )
+    scope = fields.Selection(
```
pedrobaeza, line 52: For both the user clarity and code readability, this field should be split in 2: one for the "real" scope: company or own sale, and another for where to apply it: gross sale or margin. Margin should have indicated that it's list price minus cost price.

sale_commission/models/sale_commission.py
```diff
@@ -36,6 +42,70 @@ class SaleCommission(models.Model):
     sections = fields.One2many(
         comodel_name="sale.commission.section", inverse_name="commission")
     active = fields.Boolean(default=True)
+    period = fields.Selection(
```
pedrobaeza, line 45: Why adding at commission level the settlement period? This data has sense at agent level, not at commission level

sale_commission/models/sale_commission.py
```diff
@@ -36,6 +42,70 @@ class SaleCommission(models.Model):
...
+    )
+    scope = fields.Selection(
+        selection=[("own_sales", "On his customer sales"),
+                   ("company_sales", "On company sales"),
+                   ("own_margin", "On his customers margin"),
+                   ("company_margin", "On company margin")],
+        required=True, default="own_sales",
+    )
+
+    has_company_scope = fields.Boolean(compute="_compute_scope")
```
pedrobaeza, line 60: These fields are not used and they don't make sense when splitting the other field

sale_commission/wizard/wizard_settle.py
```diff
@@ -32,35 +39,43 @@ class SaleCommissionMakeSettle(models.TransientModel):
     agents = fields.Many2many(comodel_name='res.partner',
                               domain="[('agent', '=', True)]")
 
-    def _get_period_start(self, agent, date_to):
+    @classmethod
```
pedrobaeza, line 42: Instead of @classmethod, put this as @api.model methods (it's the recommended way for methods that doesn't matter the value of self - https://www.odoo.com/documentation/8.0/reference/orm.html#openerp.api.model).

pedrobaeza:
This hasn't been work for a long time and it's conflicting everything. If you still think this is useful, please reopen a new PR without conflicts.

PR #29, opened 2015-08-28
[IMP] Automatic handling of commissions

veloutin:
Hide commissions from view in sales order and invoices. This lets the default
commissions do their work. The interface really breaks down when there are
many commissions to set, and normally there shouldn't be a need for manual
input on a per-line basis.

This is rebased on #28
This is part 3 of #27

### DIFF
--- a/sale_commission/models/sale_commission.py
+++ b/sale_commission/models/sale_commission.py
@@ -115,6 +115,32 @@ class SaleCommission(models.Model):
                 return base * section.percent / 100.0
         return 0.0
 
+    def get_default_commissions(self):
+        partner_obj = self.env['res.partner']
+        res = []
+        if self.env.context.get('partner_id'):
+            partner = partner_obj.browse(
+                self.env.context['partner_id'])
+            for agent in partner.agents:
+                if agent.commission and agent.commission.has_own_scope:
+                    comm = agent.commission
+                    res.append({
+                        'agent': agent.id,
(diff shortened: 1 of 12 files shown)


veloutin:
Rebased again on #28 to have full scope handling.

sale_commission/models/sale_commission.py
```diff
@@ -23,6 +23,12 @@
 from openerp import models, fields, api, exceptions, _
 
 
+PERIOD_MONTH = "monthly"
+PERIOD_QUARTER = "quaterly"
+PERIOD_SEMI = "semi"
+PERIOD_YEAR = "annual"
```
yvaucher, line 29: Is this use outside of the selection?
Wouldn't it be better to make a single constant that defines the whole selection list?

yvaucher, line 29: Ok I see that you used it later, you could do another constant with the full list and keep those

yvaucher:
Just a little remark otherwise :+1:

max3903:
:+1:

pedrobaeza:
This should be done via a security group, like the "Show discount in sale lines". This way, you can configure which users see the commissions and which ones don't. This configuration should be set at general level or per user level.

pedrobaeza:
This hasn't been work for a long time and it's conflicting everything. If you still think this is useful, please reopen a new PR without conflicts.

PR #31, opened 2015-09-30
Sale commission product

Alejandro-Asr:
New module for commission exception by product


sale_commission_product/README.rst
```diff
@@ -0,0 +1,32 @@
+Commission By Product
```
pedrobaeza: Can you follow current OCA's template: https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst?

sale_commission_product/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Sale commissions product',
+    'version': '1.0',
```
pedrobaeza: Use new version numbering scheme: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

sale_commission_product/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+    ],
+    'contributors': [
+        "Alejandro Sánchez <alejandro@asr-oss.com>",
+    ],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/product_template_view.xml",
+    ],
+    "installable": True,
+    "auto_install": True,
```
pedrobaeza: Don't put it auto_install. There are people that don't want this feature. auto_install is for glue modules that fix functionality when some modules are installed together.

sale_commission_product/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+{
+    'name': 'Sale commissions product',
+    'version': '1.0',
+    'author': 'Asr Oss - Alejandro Sánchez, '
+              'Odoo Community Association (OCA)',
+    "category": "Generic Modules/Sales & Purchases",
+    'license': 'AGPL-3',
+    'depends': [
+        'sale_commission',
+        'sale_stock'
```
pedrobaeza: Why sale_stock instead of product?

sale_commission_product/models/product_template.py
```diff
@@ -0,0 +1,64 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp import models, fields, api
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
```
pedrobaeza: Why on product.template and not in product.product?

pedrobaeza:
The icon includes the stock symbol, but this one is not related to stock. You should look for an icon that represents a product.

sale_commission_product/models/product_template.py
Alejandro-Asr: integrity with https://github.com/OCA/commission/blob/8.0/sale_commission/models/product_template.py

pedrobaeza: Yeah, but in that case, the field was a boolean, that it's propagated from product.template to product.product, having the best of both (set at template or at product level). In one2many, you need an inverse many2one field, that can only be of one model, so you need to step down this field to product.product.

Alejandro-Asr:
This update contains:
- Change README.rst according OCA
- Change Version according OCA
- Change Auto_install --> False
- Remove dependency sale_stock
- Change model product_template to product_product
- Change icon
- Update translation

pedrobaeza:
:+1: 

Thanks.

anajuaristi:
Tested functionally

One single concern. Steps to reproduce:
1. create a "general" commission setting
2. Create an "specific" commission for one product
3. Go to sales, create a new quotation. Include the product having specific commission on sale order line --> The "general" commision is applied so user has to select the correct one. 

There should  be a kind of "priority" for product specific commissions over general commissions where they overlap.
If correct one is selected, commission settle and invoice is correctly created. So :+1: functionally RTM
Thank you

ivantodorovich:
:+1: it seems good!

## 

ping @antwertinger since he asked for this https://github.com/OCA/commission/issues/44

gemmabr:
Works fine for me!

pedrobaeza:
@Alejandro-Asr, can you please make some things before merging this?:
- Use short headers, which is the new standard, to have smaller LOC in all modules. You can use this as template: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py
- Don't include POT file. It's generated automatically before being uploaded to Transifex, so it can be outdated.
- Use one2many_tags widget like in https://github.com/OCA/commission/pull/78

Alejandro-Asr:
This update contains:
- Use short headers.
- Delete pot file.
- Use one2many_tags widget.

pedrobaeza:
Merging this old one.

PR #34, opened 2015-10-26
[FIX] sale.commission.settlement access error opening res.partner for…

pablocm-aserti:
…m for users w/o sale rights

Users without sales rights can still open the res.partner form -eg. trough Messaging->Organizer->Contacts- but doing so with sale_commission module installed rises a "not enough rights error" for sale.commission.settlement.
This fix solves the error hiding all the commission fields to users without basic sale rights


pedrobaeza:
:+1: 

Thanks for the fix.

oihane:
:+1: LGTM

PR #38, opened 2015-10-30
8.0 sale commission multiple

Cywaithaka:
Offer support for commissions on first sale only per customer. 
Will add a boolean check to sale.commission model. Commission amount will evaluate to zero if the commission type is applicable for first purchase only and the customer has purchased before

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -22,7 +22,7 @@
 ##############################################################################
 {
     'name': 'Sales commissions',
-    'version': '2.0',
+    'version': '2.1',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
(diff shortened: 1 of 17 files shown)


pedrobaeza:
Sorry, but you're mixing here changes from Vincent Vinet, and the branch is not rebased. You should fix this and present one PR for each change.

sale_commission/__openerp__.py
```diff
@@ -22,7 +22,7 @@
 ##############################################################################
 {
     'name': 'Sales commissions',
-    'version': '2.0',
+    'version': '2.1',
```
elicoidal, line 25: Please use OCA versioning according to https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md

pedrobaeza:
Closing as no answer

PR #39, opened 2015-11-13
Created new field to specify the commission creation according to the…

rvalyi:
… state of the invoice


rvalyi:
unverified CLA email, submitting again in a minute

PR #40, opened 2015-11-13
Created new field to specify the commission creation according to the…

rvalyi:
… state of the invoice

PR #43, opened 2015-11-30
[IMP] sale_commission: base amount for computation selectable

oihane:
You can now select if commission is calculated from subtotal
or from margin

### DIFF
--- a/sale_commission/README.rst
+++ b/sale_commission/README.rst
@@ -1,3 +1,8 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=================
 Sales commissions
 =================
 
@@ -6,6 +11,9 @@ them in customers and sales orders.
 
 You can then make the settlements of these commissions, and generate the
 corresponding supplier invoices to pay their commissions fees.
+
+You can define which base amount is going to be taken into account: net amount
+(based on margin) or gross amount (line subtotal amount)
 
 Known issues / Roadmap
 ======================
@@ -35,6 +43,7 @@ Contributors
 * Giorgio Borelli <giorgio.borelli@abstract.it>
 * Daniel Campos <danielcampos@avanzosc.es>
 * Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Oihane Crucelaegui <oihanecruce@gmail.com>
 
 Icon
 ----
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -1,44 +1,28 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    Copyright (C) 2011 Pexego Sistemas Informáticos (<http://www.pexego.es>).
-#    Copyright (C) 2015 Avanzosc (<http://www.avanzosc.es>)
-#    Copyright (C) 2015 Pedro M. Baeza (<http://www.serviciosbaeza.com>)
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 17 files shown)


pedrobaeza:
Include the new possibilities on the README.rst file.

sale_commission/models/account_invoice.py
```diff
@@ -124,17 +124,23 @@ class AccountInvoiceLineAgent(models.Model):
     def onchange_agent(self):
         self.commission = self.agent.commission
 
-    @api.one
-    @api.depends('commission.commission_type', 'invoice_line.price_subtotal')
+    @api.depends('commission.commission_type', 'invoice_line.price_subtotal',
+                 'commission.amount_base_type')
     def _get_amount(self):
```
pedrobaeza: rename it to `_compute_amount` according with last standards.

sale_commission/models/sale_commission.py
```diff
@@ -40,6 +40,10 @@ class SaleCommission(models.Model):
         [('open', 'Invoice Based'),
          ('paid', 'Payment Based')], string='Invoice Status',
         required=True, default='open')
+    amount_base_type = fields.Selection(
+        selection=[('subtotal', 'Subtotal'),
```
pedrobaeza: Call it "Gross amount"

sale_commission/models/sale_commission.py
```diff
@@ -40,6 +40,10 @@ class SaleCommission(models.Model):
...
+                   ('margin', 'Margin')],
```
pedrobaeza: Call it "Net amount", and put a help explaining a bit more what is each thing.

sale_commission/models/sale_order.py
```diff
@@ -88,19 +88,24 @@ class SaleOrderLineAgent(models.Model):
...
-    @api.one
     @api.onchange('agent')
     def onchange_agent(self):
         self.commission = self.agent.commission
 
-    @api.one
-    @api.depends('commission.commission_type', 'sale_line.price_subtotal')
+    @api.depends('commission.commission_type', 'sale_line.price_subtotal',
+                 'commission.amount_base_type')
     def _get_amount(self):
```
pedrobaeza: Call it also `_compute_amount`

sale_commission/README.rst
```diff
@@ -7,6 +12,9 @@ them in customers and sales orders.
 You can then make the settlements of these commissions, and generate the
 corresponding supplier invoices to pay their commissions fees.
 
+You can define which base amount are going to take into account, net amount
```
pedrobaeza: s/are going to take into account,/is going to be taken into account:

pedrobaeza:
Thanks for the complete test suite. Definitively :+1: , but just 2 minor things:
- You can replace `(c)` by `©`.
- People from Akretion are not credited, and they should, because they developed this PR: https://github.com/OCA/commission/pull/40. @rvalyi, who we should credit?

pedrobaeza:
Ah, and please squash the commits.

oihane:
Something went really bad when making the squash...
A part of the code has disappeared... Please @pedrobaeza set again **work in progress**

pedrobaeza:
Again :+1:

Daniel-CA:
:+1:

sale_commission/models/account_invoice.py
```diff
@@ -112,39 +96,44 @@ class AccountInvoiceLineAgent(models.Model):
...
-            subtotal = self.invoice_line.price_subtotal
-            if self.commission.commission_type == 'fixed':
-                self.amount = subtotal * (self.commission.fix_qty / 100.0)
-            else:
-                self.amount = self.commission.calculate_section(subtotal)
-
-    @api.one
+    @api.depends('commission.commission_type', 'invoice_line.price_subtotal',
+                 'commission.amount_base_type')
+    def _compute_amount(self):
```
eLBati, line 112: In general, I would not change method names of stable modules, as there could be other modules using/extending them, that would suddenly stop working

pedrobaeza, line 112: This change is for following guidelines, but I agree that it can break something.

PR #47, opened 2015-12-22
[FIX] Fix singleton error

hurrinico:
I had this Problem:
-i create a SO with 2 lines; every lines has no product and 1 agent
-i validate the sale order and create the invoice
-try to validate invoice and system return ValueError Expected singleton: account.invoice.line.agent(

I create this PR in order to resolve this Problem

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -131,7 +131,7 @@ class AccountInvoiceLineAgent(models.Model):
         # Count lines of not open or paid invoices as settled for not
         # being included in settlements
         for line in self:
-            self.settled = (line.invoice.state not in ('open', 'paid') or
+            line.settled = (line.invoice.state not in ('open', 'paid') or
                             any(x.settlement.state != 'cancel'
                                 for x in line.agent_line))
 


sale_commission/models/account_invoice.py
```diff
@@ -125,15 +125,15 @@ def _compute_amount(self):
...
                     line.amount = line.commission.calculate_section(subtotal)
 
+    @api.one
     @api.depends('agent_line', 'agent_line.settlement.state', 'invoice',
                  'invoice.state')
     def _compute_settled(self):
         # Count lines of not open or paid invoices as settled for not
         # being included in settlements
-        for line in self:
-            self.settled = (line.invoice.state not in ('open', 'paid') or
```
pedrobaeza, line 134: Instead of changing to @api.one, just put here `line.settled = ...`

hurrinico, line 134: insn't better to use api.one in computed fields?

pedrobaeza, line 134: `@api.one` is deprecated in v9, so better to directly use `@api.multi`

hurrinico:
@pedrobaeza i've made the requested changes

pedrobaeza:
Thanks for the fix!

:+1: 

I fast-track because it's critical.

PR #50, opened 2016-01-05
[FIX] Add new status

hurrinico:
I've create this PR in order to solve this Problem:
I create a Sale Order, validate and create an invoice from it, validate the invoice and pay it in the future (in a different period).
If i compute settle until today the invoice will be counted even if it's not already paid

### DIFF
--- a/sale_commission/wizard/wizard_settle.py
+++ b/sale_commission/wizard/wizard_settle.py
@@ -74,6 +74,14 @@ class SaleCommissionMakeSettle(models.TransientModel):
                             agent_lines[pos].invoice.state != 'paid'):
                         pos += 1
                         continue
+                    if agent.commission.invoice_state == 'period':
+                        not_already_paid = False
+                        for move in agent_lines[pos].invoice.payment_ids:
+                            if fields.Date.today() < move.date:
+                                not_already_paid = True
+                        if not_already_paid:
+                            pos += 1
+                            continue
                     if agent_lines[pos].invoice_date > sett_to:
                         sett_from = self._get_period_start(
                             agent, agent_lines[pos].invoice_date)
--- a/sale_commission/models/sale_commission.py
+++ b/sale_commission/models/sale_commission.py
@@ -22,7 +22,8 @@ class SaleCommission(models.Model):
     active = fields.Boolean(default=True)
     invoice_state = fields.Selection(
         [('open', 'Invoice Based'),
-         ('paid', 'Payment Based')], string='Invoice Status',
+         ('paid', 'Payment Based'),
+         ('move', 'Move Based')], string='Invoice Status',
         required=True, default='open')
     amount_base_type = fields.Selection(
         selection=[('gross_amount', 'Gross Amount'),


pedrobaeza:
I don't really understand the use case adding the move state, because I think paid already covers it.

ivantodorovich:
I believe he doesn't want to settle invoices that are paid but payment was made in a date bigger than the "until date" field. At least not in that settlement.

The use case would be, for instance, settling only the invoices paid in the previous month (or before) but not including invoices that were paid in the current month (let's say.. in the very firsts days of the month, before the settlement wizard is ran).

Maybe this should be a FIX to "Payment Based" instead of adding another option, since it's actually what one would expect from this.

sale_commission/wizard/wizard_settle.py
```diff
@@ -74,6 +74,14 @@ def action_settle(self):
                             agent_lines[pos].invoice.state != 'paid'):
                         pos += 1
                         continue
+                    if agent.commission.invoice_state == 'period':
```
pedrobaeza, line 77: This should be 'move', not 'period'?

pedrobaeza:
It's still very confusing...

sale_commission/wizard/wizard_settle.py
ivantodorovich, line 77: Indeed, that should be 'move'. 
Shouldn't it, @hurrinico ?

ivantodorovich:
Lo escribo en español, porque sé que es también tu idioma principal, y creo que voy a poder explicarlo mejor.

Imaginate este escenario: 

La fecha es 10 de Febrero de 2016, y queremos liquidar las comisiones de Marcelo, nuestro agente, sobre facturas pagadas en forma mensual. 

El mes que corresponde liquidar es el mes de Enero, ya que Febrero se liquidará recién los primeros días de Marzo.

Pero resulta que a Juan, un cliente que paga siempre en término, tiene en su cuenta corriente una factura emitida el 25/01/2016, y cancelada el 05/02/2016.

No queremos que esa factura ingrese en la liquidación de Enero, ya que se pagó en Febrero. Debería ingresar recién en la próxima liquidación.

Este PR hace eso, asegurándose que no sólo la fecha de factura sea menor al 01/02/2016 (fecha que ingresaremos en el wizard) - y que dicha factura esté pagada -, sino también se asegura de que la fecha de su último pago (es decir, la fecha de cancelación) esté dentro del periodo a liquidar.

Esto, a mi parecer, es lo que uno esperaría al seleccionar 'Paid Based', es por eso que creo que los cambios sugeridos por @hurrinico , en lugar de crear una nueva opción de invoice_state, deberían ser incorporados dentro de 'paid'.

PD: Al menos así lo entiendo yo.. como siempre, puedo equivocarme :smile_cat:

pedrobaeza:
Thanks for the explanation @ivantodorovich. I continue in English because we need to make each other understandable. Well, that can be, but let's see what @hurrinico says. Indeed currently I see now a potential problem where an invoice cannot be settled, because is not paid until the next period. @oihane, can you confirm this?

hurrinico:
Hi @pedrobaeza sorry for delay but is a very busy period. 
I agree with you and @ivantodorovich i'll fix this PR leaving 2 states and applying my fix to the existing one.
I'll work on it soon (about friday i think).

pedrobaeza:
Is this going to be continued?

pedrobaeza:
Closing as no answer.

PR #53, opened 2016-01-20
[FIX] sale_commission: Refunds from invoices. Fixes #51

### LINKED ISSUE #51: Refound invoice no set agent
@pedrobaeza 
when you create a refund from a invoice does not set the corresponding agentin in lines.

captura

pedrobaeza:
When creating a refund from an invoice, pass correct values
(converted as dictionary) for the agents lines, so that the refund
keep the same commissions (but reversed) than the original invoice.

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -46,10 +46,18 @@ class AccountInvoice(models.Model):
         """ugly function to map all fields of account.invoice.line
         when creates refund invoice"""
         res = super(AccountInvoice, self)._refund_cleanup_lines(lines)
-        for line in res:
-            if 'commission_ids' in line[2]:
-                commission_ids = [(6, 0, line[2]['commission_ids'])]
-                line[2]['commission_ids'] = commission_ids
+        if lines and lines[0]._name != 'account.invoice.line':
+            return res
+        for i, line in enumerate(lines):
+            vals = res[i][2]
+            agents = super(AccountInvoice, self)._refund_cleanup_lines(
+                line['agents'])
+            # Remove old reference to source invoice
+            for agent in agents:
+                agent_vals = agent[2]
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@ivantodorovich @oihane, please check this one.

oihane:
Is it correct to do twice `super`?

pedrobaeza:
We are not making super with the same argument. Take into account that this method serves as a cleaning method. We need to clean the data in the one2many field also. Odoo already calls this method several times (for the invoice line and for the tax lines).

oihane:
After a quick test :+1:

pedrobaeza:
@ivantodorovich, can I get another positive review? Or @Daniel-CA, can you review this one?

ivantodorovich:
@pedrobaeza shouldn't it also be inverting the commission line amount? to be consistent with https://github.com/OCA/commission/pull/56

otherwise :+1: just tested it

pedrobaeza:
It's not needed, as the negative amounts are obtained on the computed field, so the fields we pass are correct.

oihane:
**I did a quick test in runbot**
- Customer invoice:
  image
- Customer refund invoice:
  image

@ivantodorovich does this solve your doubt?

Daniel-CA:
That's fine for me. :+1:

ivantodorovich:
@oihane actually, no. Even if it's showing a negative commission on the invoice total, it's only negative there, the agent line's commission is still possitive.

So, when you make that settlement, you will be paying 600 $ to your agent. (300+300) instead of 0 (300-300)

But I'll see how it works with https://github.com/OCA/commission/pull/56 and continue the work there if needed, since it's within its scope. I'm away for a few days that's why I didn't make much progress.

ivantodorovich:
I stand corrected. It works perfectly! Sorry :flushed:

PR #54, opened 2016-01-25
[8.0][sale_commission] Change visibility of agents field to only show when partner is a customer.

ivantodorovich:
(instead of showing when agent is not an agent)

In our company, an agent might also be a customer.

We think it makes more sense this way.

Comments?

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.0.0',
+    'version': '8.0.2.1.0',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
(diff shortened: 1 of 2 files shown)


sale_commission/__openerp__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.0.0',
+    'version': '8.0.3.0.0',
```
pedrobaeza: Version should be 8.0.2.1.0

pedrobaeza:
Comment about version number, but the change seems :+1: 

Travis is failing because a bad test definition that I will make a PR. You'll need then only a rebase.

ivantodorovich:
@pedrobaeza Ok! Let me know when it's done so I rebase :smile:

pedrobaeza:
You can now rebase because the branch is green.

ivantodorovich:
Done!

oihane:
LGTM :+1:

pedrobaeza:
OK, I think we can merge this.

Issue #55, opened 2016-01-25
Refund invoices settlements should be 'negative'


ivantodorovich:
Even though `account.invoice.commission_total` is correct (ie: $ -10), when settling the `sale.commission.settlement` is incorrect (positive sign --> ie: $ 10), hence paying the agent twice for something it shouldn't have been paid.

ivantodorovich:
I believe it's considered here `account.invoice._compute_commission_total`:

``` python
# Consider also purchase refunds, although not in the initial scope
if record.type in ('out_refund', 'in_refund'):
      record.commission_total = -record.commission_total
```

But nowhere else (that I can find).

`sale.commission.settlement.line.settled_ammount` doesn't hold any values, but instead relates to `account.invoice.line.agent.amount`

pedrobaeza:
You're right. The best is to put the negative on the lines, so in the settlement it appears automatically correct. The only check you should do is if the total is negative (because there are more refunds than invoices), the invoice to create should be a refund.

Can you make a PR with that?

pedrobaeza:
OK, I see now the PR, hehe. Then change according comment about the invoice creation, but I see that you have also thought about the same method :wink:

PR #56, opened 2016-01-25
[8.0][FIX][sale_commission] Fix refunds (negative commissions)

ivantodorovich:
Fixes https://github.com/OCA/commission/issues/55

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.2.1',
+    'version': '8.0.2.2.2',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
@@ -30,6 +30,7 @@
         "Daniel Campos <danielcampos@avanzosc.es>",
         "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
         "Oihane Crucelaegui <oihanecruce@gmail.com>",
+        "Iván Todorovich <ivan.todorovich@gmail.com",
     ],
     "data": [
         "security/ir.model.access.csv",
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -17,9 +17,6 @@ class AccountInvoice(models.Model):
             record.commission_total = 0.0
             for line in record.invoice_line:
                 record.commission_total += sum(x.amount for x in line.agents)
-            # Consider also purchase refunds, although not in the initial scope
-            if record.type in ('out_refund', 'in_refund'):
-                record.commission_total = -record.commission_total
 
     commission_total = fields.Float(
         string="Commissions", compute="_compute_commission_total",
@@ -132,6 +129,9 @@ class AccountInvoiceLineAgent(models.Model):
                     line.amount = subtotal * (line.commission.fix_qty / 100.0)
                 else:
                     line.amount = line.commission.calculate_section(subtotal)
+                # Refunds commissions are negative
+                if line.invoice.type in ('out_refund', 'in_refund'):
+                    line.amount = -line.amount
 
     @api.depends('agent_line', 'agent_line.settlement.state', 'invoice',
                  'invoice.state')
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -71,7 +71,8 @@ class Settlement(models.Model):
         invoice_obj = self.env['account.invoice']
         invoice_vals = {
             'partner_id': settlement.agent.id,
-            'type': 'in_invoice',
+            'type': ('in_invoice' if journal.type == 'purchase' else
+                     'in_refund'),
             'date_invoice': date,
             'journal_id': journal.id,
             'company_id': self.env.user.company_id.id,
@@ -121,15 +122,26 @@ class Settlement(models.Model):
         return []
 
     @api.multi
-    def make_invoices(self, journal, product, date=False):
+    def make_invoices(self, journal, refund_journal, product, date=False):
         invoice_obj = self.env['account.invoice']
         for settlement in self:
+            # select the proper journal according to settlement's amount
+            # considering _add_extra_invoice_lines sum of values
+            extra_invoice_lines = self._add_extra_invoice_lines(settlement)
+            extra_total = sum(x['price_unit'] for x in extra_invoice_lines)
+            invoice_journal = (journal if
+                               (settlement.total + extra_total) >= 0 else
+                               refund_journal)
             invoice_vals = self._prepare_invoice_header(
-                settlement, journal, date=date)
+                settlement, invoice_journal, date=date)
             invoice_lines_vals = []
             invoice_lines_vals.append(self._prepare_invoice_line(
                 settlement, invoice_vals, product))
-            invoice_lines_vals += self._add_extra_invoice_lines(settlement)
+            invoice_lines_vals += extra_invoice_lines
+            # invert invoice values if it's a refund
+            if invoice_vals['type'] == 'in_refund':
+                for line in invoice_lines_vals:
+                    line['price_unit'] = -line['price_unit']
             invoice_vals['invoice_line'] = [(0, 0, x)
                                             for x in invoice_lines_vals]
             invoice = invoice_obj.create(invoice_vals)
--- a/sale_commission/wizard/wizard_invoice.py
+++ b/sale_commission/wizard/wizard_invoice.py
@@ -11,7 +11,11 @@ class SaleCommissionMakeInvoice(models.TransientModel):
 
     def _default_journal(self):
         return self.env['account.journal'].search(
-            [('type', '=', 'purchase')])[0]
+            [('type', '=', 'purchase')])[:1]
+
+    def _default_refund_journal(self):
+        return self.env['account.journal'].search(
+            [('type', '=', 'purchase_refund')])[:1]
 
     def _default_settlements(self):
         return self.env.context.get('settlement_ids', [])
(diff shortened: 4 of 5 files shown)


ivantodorovich:
I didn't modify the version because I've made another PR that also modifies it.

What should I do in these cases? Give both PR the same version (current+1)?

pedrobaeza:
You can raise also the version number here, because you don't know which PR will be merge before. And when one of them is merged, then rebase and change the number accordingly.

sale_commission/models/account_invoice.py
```diff
@@ -124,6 +121,10 @@ def _compute_amount(self):
                     line.amount = subtotal * (line.commission.fix_qty / 100.0)
                 else:
                     line.amount = line.commission.calculate_section(subtotal)
+                # Refunds commissions are negative
+                # Consider also purchase refunds, although not in the initial scope
```
pedrobaeza: You can remove the second line of the comment, because is now out of scope

ivantodorovich:
@pedrobaeza , in response to your comment on https://github.com/OCA/commission/issues/55:

> You're right. The best is to put the negative on the lines, so in the settlement it appears automatically correct. The only check you should do is if the total is negative (because there are more refunds than invoices), **the invoice to create should be a refund**.
> 
> Can you make a PR with that?

I'm actually still learning and getting familiarized with odoo. I haven't messed with invoices and accounts yet.

For what I can tell, it isn't enough to just change `'type'` to `'in_refund'`, isn't it? 
Should I also select a proper journal of type `'purchase_refund'`?

If so, it might be a somewhat complicated change, since the wizard asks for the journal before knowing if the settlement is possitive or negative.

We could also ask for a refund journal in the wizard, and then decide.  (and change settlement's `def make_invoices(self, journal, product, date=False)` to accept the second journal)

Or we could, maybe, add a configuration page, and set the journals there... and avoid asking for journals (and product) in the wizard entirely. (Also avoiding passing those arguments around)

I could use some guidance here :flushed:

pedrobaeza:
We need both the type field and the journal, so you're right, the journal is asked before knowing the amount currently. The best is to include both journals as not required, and claim when the corresponding type is not filled if needed. What do you think?

ivantodorovich:
You mean throwing an error when the journal is needed and missing?
Why not require both journals from the start?

Regarding the configuration page. Wouldn't it be usefull to predefine the wizard's journals and product in a config page? 
Maybe it's a topic for another PR, but it's kinda related since it could be possible to make the wizard autoselect both jorunals by default without being error-prone (_ie: user forgetting to correctly select second journal since it's almost never used_)

pedrobaeza:
About the config page, we don't need another configuration hide in Settings menu. This should be more direct in the wizard, but I agree we can put by default the first found of each type, and put them as required.

ivantodorovich:
Ok, I'll make them both required then.

About the config page, I agree, there are many configuration pages already. But still, it could be usefull to set predefined journals somewhere, instead of just using the first one found. 
In our case, for instance, the first journal isn't the preferred one. The user has to manually change it every time. Same happens with the product, where there's no default. 
There must be a place to set things like this... it's a shame we lose features just because the settings menu/ui is cluttered.

ivantodorovich:
On a side note, messing with the module on my testing database, I found out this module doesn't behave with multi-company very well. (_at least to my poor understanding_)

It allows to select journals from another company (which is good, actually) but then the invoices created don't respect the journal's company. Instead, it uses `self.env.user.company_id.id`. Shouldn't invoices' company be the same as their journal's? If so, I could easily fix that in this same PR.

pedrobaeza:
Iván, you can already predefine another journal. You just need to play with the ir.values table, that allows to predefine a value in a model. It's standard everywhere. In debug mode, you also have the possibility to add this from the dropdown list that appears.

About multi-company, the module is not totally multi-company aware (it's said on the README), but that question is because you're using admin user, what you mustn't do ever in a multi-company environment. In a regular user, record rules limit the selectable journals to the ones that belong to the company.

ivantodorovich:
@pedrobaeza thank you for the clarification and for your patience. :blush: 
Every day odoo's power amazes me a little bit more.

pedrobaeza:
Thanks for your contributions. That's the open source spirit!

ivantodorovich:
@pedrobaeza if you don't mind provinding some more guidance :flushed: ... 

In order to create a proper refund invoice, I need to invert invoice lines values. Since in refunds, positive is negative. Right?

Should I also invert the lines added by the hook `_add_extra_invoice_lines` ?
What's the use case of this method? I mean, why would anyone add lines to the invoice after each settlement line? (asuming, in the future, we group invoices by agents)

pedrobaeza:
No, you don't need to invert. The amount is in positive. Odoo takes care of the type for creating correctly the account moves. That hook is for possible extensions, but they are not used for now.

ivantodorovich:
@pedrobaeza when you have some time, can you please check the changes?

I've inverted invoice lines since otherwise it wouldn't work (settlements are negative, remember? -- in refunds, they must be positive)

I've also inverted _add_extra_invoice_lines's lines, because I think if some other module uses this hook, it wouldn't know if the created invoice is a refund or not. I've also considered those lines' amount when to select the journal.

I've tested it and it seems to work as expected.. but I'm not very happy with the code, seems a bit messy.

What do you think?

sale_commission/wizard/wizard_invoice.py
```diff
@@ -13,6 +13,10 @@ def _default_journal(self):
         return self.env['account.journal'].search(
             [('type', '=', 'purchase')])[0]
 
+    def _default_refund_journal(self):
+        return self.env['account.journal'].search(
+            [('type', '=', 'purchase_refund')])[0]
```
pedrobaeza: You can put `[:1]` instead of `[0]` to avoid an error in case there's no journal on the company (also for the other journal).

sale_commission/models/settlement.py
```diff
@@ -77,6 +77,9 @@ def _prepare_invoice_header(self, settlement, journal, date=False):
             'company_id': self.env.user.company_id.id,
             'state': 'draft',
         }
+        # Select proper invoice.type according to journal.type
```
pedrobaeza: You can make it one-liner in the dictionary assignation with:

```
            'type': 'in_invoice' if journal.type == 'purchase' else 'in_refund'
```

sale_commission/__openerp__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.1.1',
+    'version': '8.0.2.2.1',
```
pedrobaeza: You will have to rebase and change this number accordingly

sale_commission/models/settlement.py
```diff
@@ -121,15 +124,30 @@ def _add_extra_invoice_lines(self, settlement):
...
 
     @api.multi
-    def make_invoices(self, journal, product, date=False):
+    def make_invoices(self, journal, refund_journal, product, date=False):
         invoice_obj = self.env['account.invoice']
         for settlement in self:
+            # select the proper journal according to settlement's amount
+            # considering _add_extra_invoice_lines sum of values
+            extra_invoice_lines = self._add_extra_invoice_lines(settlement)
+            extra_total = 0
```
pedrobaeza: You can make it one line:

```
extra_total = sum(x['price_unit'] for x in extra_invoice_lines)
```

sale_commission/models/settlement.py
```diff
@@ -121,15 +124,30 @@ def _add_extra_invoice_lines(self, settlement):
...
+            for extraline in extra_invoice_lines:
+                extra_total += extra_invoice_lines['price_unit']
+            invoice_journal =   (journal
```
pedrobaeza: For making this happy for Travis it should be:

```
            invoice_journal =   (journal if 
                                 (settlement.total + extra_total) >= 0 else
                                 refund_journal)
```

sale_commission/models/settlement.py
```diff
@@ -71,7 +71,8 @@ def _prepare_invoice_header(self, settlement, journal, date=False):
         invoice_obj = self.env['account.invoice']
         invoice_vals = {
             'partner_id': settlement.agent.id,
-            'type': 'in_invoice',
+            'type': 'in_invoice' if journal.type == 'purchase'
+                                    else 'in_refund',
```
oihane:
``` python
            'type': ('in_invoice' if journal.type == 'purchase' else
                     'in_refund'),
```

sale_commission/models/settlement.py
```diff
@@ -121,15 +122,26 @@ def _add_extra_invoice_lines(self, settlement):
...
+    def make_invoices(self, journal, refund_journal, product, date=False):
         invoice_obj = self.env['account.invoice']
         for settlement in self:
+            # select the proper journal according to settlement's amount
+            # considering _add_extra_invoice_lines sum of values
+            extra_invoice_lines = self._add_extra_invoice_lines(settlement)
+            extra_total = sum(x['price_unit'] for x in extra_invoice_lines)
+            invoice_journal = (journal if
+                                (settlement.total + extra_total) >= 0 else
+                                refund_journal)
```
oihane: Lines 133 and 134 had an extra space, erase it, and travis will say OK :wink:

pedrobaeza: And please squash commits in one, although your commits messages are very funny :smiley:

oihane: And you should also do a rebase... and update the version number :laughing:

ivantodorovich:
Thank you very much @pedrobaeza and @oihane :bow: :sweat_smile:

pedrobaeza:
:+1:  Merging!

Thank you to you.

PR #57, opened 2016-01-26
[FIX] sale_commission: Fix tests

pedrobaeza:
Tests create invoices in a month previous the current date, which
makes that they fail on January, as previous fiscal year doesn't
exist. This is fixed creating invoices in the current date, and
calling the settle wizard with one month more. This won't fail
on December, because there's no need of the next fiscal year for
this operation


pedrobaeza:
I fast-track this fix for getting the branch green

PR #58, opened 2016-01-26
[IMP][8.0][sale_commission] Show results on wizard

ivantodorovich:
This is a WIP.

If you like it, I'll also change `wizard_invoice`

What do you think so far?

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.1.1',
+    'version': '8.0.2.2.1',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
--- a/sale_commission/views/settlement_view.xml
+++ b/sale_commission/views/settlement_view.xml
@@ -6,7 +6,7 @@
             <field name="name">Settlements tree</field>
             <field name="model">sale.commission.settlement</field>
             <field name="arch" type="xml">
-                <tree string="Settlements"
+                <tree string="Settlements" create="false"
                       colors="red: state == 'cancel'; gray: state == 'invoiced'">
                     <field name="agent"/>
                     <field name="date_from"/>
@@ -42,7 +42,7 @@
             <field name="name">Settlements</field>
             <field name="model">sale.commission.settlement</field>
             <field name="arch" type="xml">
-                <form string="Settlement">
+                <form string="Settlement" create="false">
                     <header>
                         <button string="Make invoice"
                                 attrs="{'invisible': ['|', ('state', '!=', 'settled'), ('agent_type', '!=', 'agent')]}"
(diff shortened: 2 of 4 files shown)


pedrobaeza:
Instead of continuing in a TransientModel, you can return an action window dictionary that opens the results in normal form (without popup), and directly with the corresponding model (sale.commission.settlement)

ivantodorovich:
@pedrobaeza, take a look now.

I don't know how to remove the breadcrumbs history, though:
sin titulo

It will always append "Created Settlements" after the current navigation history. 
Is there a way to clear it?

pedrobaeza:
I'm afraid not, but that's not bad. You get a lot because you're invoking again and again the wizard, but in reality, it will be invoked only once.

ivantodorovich:
What's bad is that it will always append after the current history. In my previous picture: "Clientes" (Customers)

**This is bad:**

sin titulo1

sin titulo

**Possible solution:**

I found out something, but it isn't ideal either. We can remove the whole breadcrumbs and title adding 

``` python
'flags': {'display_title': False}
```

Action buttons, search bar and everything else will still show. Only the title/breadcrumbs is removed.

sin titulo4

I'm more inclined for this option. Even though we lose the "Created Settlements" title, at least it's not misleading some false breadcrumbs.

What do you think?

pedrobaeza:
I don't think that's bad. It happens continuously. Try for example to chart of accounts in accounting. But it's consistent, so I think you should keep it as is.

ivantodorovich:
Ok, then. My neurotic mind is not happy, but I guess this is the way Odoo SA does it, so it's consistent!

I've used the same approach on wizard_invoice. It's now ready for review.

On a side note:
Do they accept PRs? Or is it a waste of time?
It would be very easy to add a new act_window option to clear the breadcrumbs.
When searching the web for this, I found a few other people with the same concern.

pedrobaeza:
You can try, but I don't think they will accept it (at least on a stable branch)

pedrobaeza:
You shouldn't remove the possibility to create manually settlements or settlements lines, for any adjustments you want to make manually (or even extra payments to your agents).

ivantodorovich:
@pedrobaeza wouldn't you just use supplier invoices for that purpose?

Settlement's lines are readonly, so you can't really add or adjust lines manually.

Even if they weren't readonly, it wouldn't be possible to add a custom line, since they are always related to the agent's lines. So any adjustment should really be made from the invoice agent line before the wizard is ran.

I don't think settlements are meant to be manually created nor edited. I think they are meant to be created only from the wizard. That way its integrity is guaranteed.

Previous to this PR, settlement's views (with create buttons) weren't accessible from anywhere in the menu. Since now I'm exposing them, I think it's safe to remove the create button, in order to maintain settlements integrity.

pedrobaeza:
OK, you're right, let's put it this way.

Can you please squash the commits in one, as they're very little or fixes to your previous code? 

:+1: when done

ivantodorovich:
Yes of course. :+1:

pedrobaeza:
@oihane, please review

oihane:
Code checked, looks good :+1:

Issue #63, opened 2016-01-30
Creating agent from within the invoice line agent line widget doesn't preset partner.agent to true.


ivantodorovich:
Creating agent from within the invoice line agent line widget doesn't preset partner.agent to true.

pedrobaeza:
This can be solved setting in the context of the field `{'default_agent': True, 'default_customer': False, 'default_supplier': True}`

PR #64, opened 2016-01-31
[FIX][8.0][sale_commission] Fixes #61 and #63.

### LINKED ISSUE #61: [sale_commission] Commissions are not being settled for invoices duplicated from a settled invoice.
1. Create an invoice with agent lines
2. Settle the invoice from the commissions wizard
3. Duplicate the invoice
4. Try to settle de invoice

Result: Invoice is ignored in the settlement.
Expected: Invoice commission should be settled.

This bug won't occur, though, if the invoice is duplicated from a not settled invoice.

I'm guessing the `account.invoice.line.agent`'s `settled` field is being copied aswell when duplicating an invoice, hence ignoring this invoice line from settlements.

Is there a way to avoid this? a way to reset this field value when duplicating?
I don't know how to do it, yet, at least not without investigating the web for a few hours.

ivantodorovich:
Fixes #61 and #63.

Also I've reordered field definitions for readability.

Even though it wasn't needed to fix the bugs, I've added `Copy=False` in `invoice_line` field. I'm guessing it was being copied and then replaced, so it was ok. But I think it's better -- safer -- if it's not copied in the first place.

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.2.2',
+    'version': '8.0.2.3.0',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -86,27 +86,37 @@ class AccountInvoiceLineAgent(models.Model):
     _name = "account.invoice.line.agent"
 
     invoice_line = fields.Many2one(
-        comodel_name="account.invoice.line", required=True, ondelete="cascade")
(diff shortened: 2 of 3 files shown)


sale_commission/__openerp__.py
```diff
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.2.2',
+    'version': '8.0.2.2.3',
```
pedrobaeza: This should be 8.0.2.3.0 as a module update is needed (https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers)

ivantodorovich:
@pedrobaeza done! :smile:

pedrobaeza:
:+1:

oihane:
:+1: LGTM

Issue #62, opened 2016-01-30
Error when editing invoice lines


ivantodorovich:
I'm getting this error when **editing** invoice lines.

**Steps to reproduce:**
1. Create an invoice with commissions
2. Save it (but don't validate)
3. Open and Edit the invoice
4. Edit some invoice line quantity or product. Maybe also click the invoice line agent lines.
5. Try to save the line (lose focus)

This error window raises 3 times:

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 537, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 574, in dispatch
    result = self._call_function(**self.params)
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 310, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/service/model.py", line 118, in wrapper
    return f(dbname, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 307, in checked_call
    return self.endpoint(*a, **kw)
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 803, in __call__
    return self.method(*args, **kw)
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/http.py", line 403, in response_wrap
    response = f(*args, **kw)
  File "/home/itodorovich/odoo/server/addons/web/controllers/main.py", line 944, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/itodorovich/odoo/server/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/api.py", line 256, in wrapper
    return old_api(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/odoo-8.0-py2.7.egg/openerp/api.py", line 387, in old_api
[...]

Issue #52, opened 2016-01-12
TypeError: Cannot read property 'constructor' of undefined


odortega:
![sale_commission_error](https://cloud.githubusercontent.com/assets/4130995/12280025/0b3970a6-b959-11e5-9dfc-251e3631182b.png)
I am using "sale_commission" module for manage sales commission to sales team https://www.odoo.com/apps/modules/8.0/sale_commission/, when I am creating a sale order all work fine till I try to add sale order line whith a different product than before line.

May you help me?
Thanks very much : ]

Odoo V: 8.0.2

This is the trace:

TypeError: Cannot read property 'constructor' of undefined
    at wrap_context (https://moogo.co/web/js/web.assets_backend/1efc6a0:2918:7)
    at https://moogo.co/web/js/web.assets_backend/1efc6a0:2921:127
    at Array.reduce (native)
    at Function._.reduce._.foldl._.inject (https://moogo.co/web/js/web.assets_common/87b133b:11:564)
    at _.(anonymous function) [as reduce](https://moogo.co/web/js/web.assets_common/87b133b:48:939)
    at eval_contexts (https://moogo.co/web/js/web.assets_backend/1efc6a0:2919:175)
    at https://moogo.co/web/js/web.assets_backend/1efc6a0:2921:253
    at Array.reduce (native)
    at Function._.reduce._.foldl._.inject (https://moogo.co/web/js/web.assets_common/87b133b:11:564)
    at _.(anonymous function) [as reduce](https://moogo.co/web/js/web.assets_common/87b133b:48:939)
[...]

Issue #49, opened 2016-01-04
error in form view sales_order with unit of measure


MiguelMachadoM:
@pedrobaeza 
When add 3th line in a sale quotation I recibe following errors:

![commision1](https://cloud.githubusercontent.com/assets/7348561/12086197/0655f33a-b2c8-11e5-9b80-c3afedc205cc.PNG)

And next when try again:

![commision2](https://cloud.githubusercontent.com/assets/7348561/12086205/14af2258-b2c8-11e5-98f7-346a6653bc6c.PNG)

I've tried uninstalling other modules that affected the view and the error still occurs.

More info: if I remove Agent form customer sheet and create a new quotation this error does not occur.

P.D. to reproduce this error you must add at least 3 lines in the quotation

ivantodorovich:
I've also seen this error. It might be related to https://github.com/OCA/commission/issues/62

ivantodorovich:
https://github.com/OCA/commission/pull/78 will probably solve it.

Daniel-CA:
Tested in a runbot instance and the bug doesn't happen, can be closed.

Issue #25, opened 2015-07-09
Invoice line view becomes 'messy' when "Agents & Commissions" clicked


mfzoutsource:
Hi @pedrobaeza, FYI, my invoice line view become like this (image attached) in draft invoice (from sale process), as standard practice for me when I want to confirm some invoices, I need to make sure the information is correct. So when I clicked on 'Agents & Commissions' columns in invoice line, the view become quite messy like this. 
Thank you.

![tree_view_invoice_commission](https://cloud.githubusercontent.com/assets/13113545/8587318/5a9253cc-262c-11e5-8e99-2e6d05654a17.png)

pedrobaeza:
Yeah, that's the price we have to pay for not having the line in pop-up. Although not pretty, it's still usable. That's why I kept it. Any alternative suggestion?

ivantodorovich:
Related https://github.com/OCA/commission/issues/62

ivantodorovich:
https://github.com/OCA/commission/pull/78 will probably solve it.

ivantodorovich:
This is fixed now, since https://github.com/OCA/commission/pull/78 was merged.
@pedrobaeza or @mfzoutsource  can you close it?

PR #70, opened 2016-02-09
[WIP][8.0][sale_commission] Add security group 'Commissions in Sale Order Lines'

ivantodorovich:
Adds a security group to show/hide the agents & commissions in sale order lines and invoices.

This is related to: https://github.com/OCA/commission/issues/62, since I'm trying to avoid that error.
Also somewhat related to #49 & #25 & #52

And also inspired by this PR https://github.com/OCA/commission/pull/29, specifically this https://github.com/OCA/commission/pull/29#issuecomment-136230386 by @pedrobaeza

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.3.0',
+    'version': '8.0.2.4.0',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
@@ -34,6 +34,7 @@
     ],
     "data": [
         "security/ir.model.access.csv",
+        "security/sale_commission_security.xml",
         "views/product_template_view.xml",
         "views/res_partner_view.xml",
         "views/sale_commission_view.xml",
--- a/sale_commission/models/sale_order.py
+++ b/sale_commission/models/sale_order.py
@@ -39,9 +39,11 @@ class SaleOrderLine(models.Model):
     agents = fields.One2many(
         string="Agents & commissions",
         comodel_name='sale.order.line.agent', inverse_name='sale_line',
+        groups="sale_commission.group_commission_per_so_line",
         copy=True, readonly=True, default=_default_agents)
     commission_free = fields.Boolean(
         string="Comm. free", related="product_id.commission_free",
+        groups="sale_commission.group_commission_per_so_line",
         store=True, readonly=True)
 
     @api.model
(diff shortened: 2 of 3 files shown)


ivantodorovich:
I need some help, though. Because it's not working as expected.

If the field is hidden (user is not in the security group), the `_default_agents` is being called but returning an empty list instead of the default agents.

This is because the `context` defined in the view (`{'partner_id': partner_id}`) does not make it to `_default_agents` when the model api tries to populate defaults.

What would be the proper way to handle this?
Should I try to find the `partner_id` by browsing the `sale_order_id` of the sale order line?
Or is there a way to properly passthrough the context defined in the view?

pedrobaeza:
Well, I think you'll need a workaround for this, because if the field is not in the view, the default is not retrieved, so I think this also doesn't serve if you put the `groups` attribute in the field, not in the view (but you can try), so you need to:
- Add the field
- Add the group
- Have a boolean computed field (not stored) that says if the user belongs to the group.
- Put `attrs` attribute in the field that makes invisible or not depending on the previous compute field.

ivantodorovich:
Sadly, just making the field invisible does not avoid the errors that the widget generates (https://github.com/OCA/commission/issues/62)

Also, the defaults _are_ being retrived by model's method `create()`, but the context's attribute `partner_id` defined in the view isn't present, so `_default_agents()` returns nothing.

I could, however, add a workaround by manually setting agent's defaults in `create`, by getting the `partner_id` from browsing `order_id`. I'm still looking if there's a cleaner way, though.

ivantodorovich:
Also I realized this may bring up some unexpected behaviours.

For instance, take the following scenario: 
- User does not belong to this group (hence expecting commissions to be automatically calculated)
- Create Sale Order for customer A, add some lines.
- Save the Order
- Edit the Order
- Change the customer to Customer B
- Add more lines & save.

Some of the lines will belong to Customer A's Agent, and some to Customer B's Agent.
Expected behaviour should be that all lines belong to Customer B's Agent.

This already happens currently, but since the user can edit the lines it isn't a real problem.

I could overwrite all agent lines with defaults when saving the `sale.order` if the user doesn't belongs to the security group, but that seems also hacky and prone to another unexpected behaviour (overwriting another user explicitly set line agents)

---

I need commissions to be managed silently and consistently.
I do not need multiple agents per customer.
I do not need agents to be set for each `sale.order.line`. Just setting an agent per `sale.order` is enough.

I'm wondering if the way to go is to _hide_ features from the module with security groups, or instead fork it and remove them completely. :/

Also considering refactoring the module into two modules, in order to get a simplified version that can be published for others to use.

What would you do?

pedrobaeza:
I only see as solution to catch create and write events, but this is not a general solution for everyone.

PR #72, opened 2016-02-10
[UPD][9.0] sale_commission

oihane:


pedrobaeza:
As this is only to bring latest changes from 8.0 and all is green, I merge.

PR #75, opened 2016-02-24
[IMP]sale_commission: Added multi-company.

jesusVMayor:
Added multi-company in sale_commission and sale_stock_commission modules.

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -122,6 +122,8 @@ class AccountInvoiceLineAgent(models.Model):
     settled = fields.Boolean(
         compute="_compute_settled",
         store=True, copy=False)
+    company_id = fields.Many2one("res.company", "Company",
+                                 related="invoice.company_id", store=True)
 
     @api.multi
     def name_get(self):
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -34,6 +34,7 @@ class Settlement(models.Model):
     currency_id = fields.Many2one(
         comodel_name='res.currency', readonly=True,
         default=_default_currency)
+    company_id = fields.Many2one('res.company', 'Company')
 
     @api.depends('lines', 'lines.settled_amount')
     def _compute_total(self):
@@ -173,3 +174,5 @@ class SettlementLine(models.Model):
         related="agent_line.amount", readonly=True, store=True)
     commission = fields.Many2one(
         comodel_name="sale.commission", related="agent_line.commission")
+    company_id = fields.Many2one('res.company', 'Company',
+                                 related="settlement.company_id", store=True)
--- a/sale_commission/wizard/wizard_settle.py
+++ b/sale_commission/wizard/wizard_settle.py
@@ -66,31 +66,41 @@ class SaleCommissionMakeSettle(models.TransientModel):
                 [('invoice_date', '<', date_to_agent),
                  ('agent', '=', agent.id),
                  ('settled', '=', False)], order='invoice_date')
-            if agent_lines:
-                pos = 0
-                sett_to = fields.Date.to_string(date(year=1900, month=1,
-                                                     day=1))
-                while pos < len(agent_lines):
-                    if (agent.commission.invoice_state == 'paid' and
-                            agent_lines[pos].invoice.state != 'paid'):
-                        pos += 1
-                        continue
-                    if agent_lines[pos].invoice_date > sett_to:
-                        sett_from = self._get_period_start(
-                            agent, agent_lines[pos].invoice_date)
-                        sett_to = fields.Date.to_string(
-                            self._get_next_period_date(agent, sett_from) -
-                            timedelta(days=1))
-                        sett_from = fields.Date.to_string(sett_from)
-                        settlement = settlement_obj.create(
-                            {'agent': agent.id,
-                             'date_from': sett_from,
-                             'date_to': sett_to})
-                        settlement_ids.append(settlement.id)
-                    settlement_line_obj.create(
-                        {'settlement': settlement.id,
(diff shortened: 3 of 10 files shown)


sale_commission/models/account_invoice.py
```diff
@@ -117,6 +119,7 @@ class AccountInvoiceLineAgent(models.Model):
     settled = fields.Boolean(
         compute="_compute_settled",
         store=True, copy=False)
+    company_id = fields.Many2one('res.company', 'Company')
```
pedrobaeza: An invoice already belongs to a company, so there's no need to add it here

jesusVMayor: It is neccesary for the sale.commission.make.settle wizard

pedrobaeza: No, you should take another approach: filter first the invoices that are for that company, and then select the lines.

omar7r:
:+1:

pedrobaeza:
You have to rebase.

jesusVMayor:
rebased

sale_commission/wizard/wizard_settle.py
```diff
@@ -85,11 +85,15 @@ def action_settle(self):
                         settlement = settlement_obj.create(
                             {'agent': agent.id,
                              'date_from': sett_from,
-                             'date_to': sett_to})
+                             'date_to': sett_to,
+                             'company_id':
+                                agent_lines[pos].invoice_line.company_id.id})
```
pedrobaeza: This doesn't serve. You can have access to more than one company with your user, so your invoices will be mixed. You have to group invoices by company, and make one settlement per group.

omar7r: I think that is not necessary, it is the same problem that invoice by picking, the user must stay in the company that try to invoice.

pedrobaeza: There are multiple valid uses in multi-company environment outside admin. For example, parent companies with access to several child companies, so your code must be aware to this. 

There's already enough code in Odoo core that doesn't take this into account to add more.

omar7r: Imho this behavior is the correct the users in parent companies should be
for consulting and reporting because if we allow these users can create
settlements the normal users can not access to it by company restrictions.
At most we can check that the settlement don't mix companies.

pedrobaeza: But the code for that is the same as the code for grouping by companies. If you don't want to do it, let me propose a PR to your branch, but it's a pity that we waste this occasion to make this real multi-company aware.

I know (and this is the advice I give), that companies that are parent of another (aka company groups) are only for viewing normally, but for example in a customer I have with a franchise model, this way of doing (using a company group) is the only way they have for invoicing for example franchises. For now, I'm not going to use commissions on them, but if so, we can let the work done.

jesusVMayor:
- Modified make settle wizard.
- Modified multicompany rule of sale.commission model to allow access to commissions of parent companies.

ivantodorovich:
Haven't tested it, but code looks good to me :+1:

sale_commission/models/settlement.py
```diff
@@ -173,3 +174,4 @@ class SettlementLine(models.Model):
         related="agent_line.amount", readonly=True, store=True)
     commission = fields.Many2one(
         comodel_name="sale.commission", related="agent_line.commission")
+    company_id = fields.Many2one('res.company', 'Company')
```
pedrobaeza: This should be a stored related field to `'settlement.company_id'`

sale_commission/wizard/wizard_settle.py
```diff
@@ -66,31 +66,42 @@ def action_settle(self):
...
+                                     'date_from': sett_from,
+                                     'date_to': sett_to,
+                                     'company_id': company.id})
+                                settlement_ids.append(settlement.id)
+                            settlement_line_obj.create(
+                                {'settlement': settlement.id,
+                                 'agent_line': [(6, 0,
+                                                 [agent_lines_company[pos].id])
+                                                ],
+                                 'company_id': company.id})
```
pedrobaeza: You don't need to set this value when making it a related field

pedrobaeza:
Several things before merging:
- Some comments put inline.
- Increase the version number on the manifest to 8.0.3.0.0.
- Remove the line in the Known issues about the support for multi-company.
- You are not adding the field `company_id` to the object account.invoice.line.agent. It should be a related field to `invoice.company_id`.
- Please switch the author `Serv. Tecnol. Avanzados...` to the new company `Tecnativa`.
- Include a pre-migration script that:
  - Creates the column company_id as integer on the tables _sale_commission_ and _sale_commission_settlement_.
  - For the table _sale_commission_, you don't need to make anything else, as the existing commissions should remain visible for all companies with the empty value.
  - For the table _sale_commission_settlement_, you should perform this query to bring the correct company:
    
    ```
    UPDATE sale_commission_settlement scs
    SET company_id = res_partner.company_id
    WHERE res_partner.id = scs.agent
    ```

jesusVMayor:
Thanks for the suggestions

> You are not adding the field company_id to the object account.invoice.line.agent. It should be a related field to invoice.company_id.

Why?. There is already a multicompany rule for account.invoice.line.agent based on the company of account.invoice.line

> This should be a stored related field to 'settlement.company_id'

And the same question for settlement.line why make a related field instead of a multicomapy rule based on settlement.company_id

pedrobaeza:
The main problem with your approach is the performance. Each time a record is accessed, you need to read 2 tables for having the access. If you make a related stored field on the table, the access is done in one table, and data is even cached. The only penalty is the overhead on writing the record, but it's very short I think.

Please address the rest of the comments anyway.

pedrobaeza:
:+1: 

Please put a comment next time, because we are not notified if you squash your commits.

Issue #76, opened 2016-03-02
[BUG?] Editing the commision will recalculate all previous agent lines and settlements.


ivantodorovich:
Editing a `sale.commission` that's already in use will trigger an update on every existing invoice, sale order, and settlements that were using that commission "rule".

This is because of `@api.depends`:

``` py
    @api.depends('commission.commission_type', 'invoice_line.price_subtotal',
                 'commission.amount_base_type')
    def _compute_amount(self):
```

I think this could cause consistency errors.

IMO, settled invoices's commissions (agent lines) should remain unchanged, even if the `sale.commission` percentage / formula / rule changes.

One could argue that it shouldn't even change on unsettled invoices / sale orders. IE: If an agent renegotiates it's commission percentage, the change should apply on new sales, not previous ones.. even if they are not settled yet.

What do you think?

Should we remove `'commission.commission_type'` and `'commission.amount_base_type'` from the `@api.depends`?

pedrobaeza:
Yeah, you're totally right about that depends. Please make a PR removing it.

PR #77, opened 2016-03-02
[FIX][8.0][sale_commission] https://github.com/OCA/commission/issues/76

ivantodorovich:
https://github.com/OCA/commission/issues/76

This partially fixes the bug.

It no longer recalculates agent lines when updating a commission rule.

But, the following known issues remain:
1. It will still recalculate agent lines when making the settlement. It seems related fields trigger recomputation of fields. But, at least, it will **not** trigger updates on already created settlements.
2. When making an invoice from a sale order, it will recalculate the agent line (since it's actually a new one) If the commission rule has been updated after the sales order but before the invoice, the new rule will apply. (I don't believe this is expected, so it should be warned)

Given **1.** ; **2.** is irrelevant.. but still, I thought it should be mentioned.

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.2.3.0',
+    'version': '8.0.2.3.1',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Please change version number (the third digit, as this only requires to restart service).

ivantodorovich:
Sorry, I forgot about that.
Done!

pedrobaeza:
:+1: 

This one can be fast-tracked as this is a critical bug, so I merge.

Issue #71, opened 2016-02-09
Some terms are not being translated


ivantodorovich:
Some terms like `Net Amount`, `Gross Amount` are not being translated. 
They are not in the `.pot` file.

I've tried exporting the language file from `Settings -> Translations -> Import / Export -> Export Translation` but it seems it also adds some terms that are not of this module, so I don't know if it's reliable.

What's the proper way to generate and/or update them?

ivantodorovich:
FIXED by adding support to Transifex. It automagically generates and updates `.pot` files.

PR #79, opened 2016-03-15
[8.0] TRANSIFEX Support

ivantodorovich:
Ok, now that I learned this, I can fix https://github.com/OCA/commission/issues/71
I mean, TRANSIFEX will take care of updating the `.pot` file, right?


pedrobaeza:
:+1: Fast-tracking it

PR #80, opened 2016-03-15
[9.0] TRANSIFEX Support

ivantodorovich:
Same as https://github.com/OCA/commission/pull/79


pedrobaeza:
:+1:

As everything is green, I also fast track

PR #84, opened 2016-04-07
Commission by partner

javierjcf:
Let agents and commissions by partner and multicompany

This pull request contains the changes in the pull request:
https://github.com/OCA/commission/pull/75


pedrobaeza:
Javier, thanks for this feature, but I think that having this in a separate module makes that you need to overwrite a lot of code. Why don't you integrate this feature in the main module, having a one2many list like yours with the agent and the commission, but having a check marked by default indicating "default commission", and when you uncheck it, then you're able to modify the commission?

javierjcf:
Ok Pedro, no problem. I'm waiting for the approve of the multicompany pull request to do it more cleanly:
https://github.com/OCA/commission/pull/75

PR #89, opened 2016-05-03
Invoice from picking adding agents

aitorbouzas:
This is my first pull, sorry if i'm making any mistake.

Just updated the module in 8.0 version to be able to pass agents to the invoice from the picking. This depends on module stock_account (if i'm not mistaken).


pedrobaeza:
You already have this in https://github.com/OCA/commission/tree/8.0/sale_stock_commission.

aitorbouzas:
Oops! Sorry

PR #96, opened 2016-06-26
Commission partner

rubencabrera:
Adds the partner_id field so it can be displayed on commission lines information and used in report views (that's the only reason it's a stored field). 

After a while using sale_commission and sale_commission_stock, a couple of clients have gone mad trying to share commission information with agents, who usually work with just the partner sum of invoiced commissions. Also added the  "subtotal" field, related to the amount of the invoice line for the same reason.

### DIFF
--- a/sale_commission_partner/README.rst
+++ b/sale_commission_partner/README.rst
@@ -1,0 +1,38 @@
+Commission partner details
+===========================
+
+This module adds the partner field to sale_commision information to keep track
+of the partner related to each commissioned invoice, trying to adapt to an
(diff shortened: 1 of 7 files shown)


rubencabrera:
@bisnesmart a mí la guardia

ivantodorovich:
Shouldn't it be an update to the current module, instead of creating a new one?

rubencabrera:
@ivantodorovich That's ok with me, just did it in a new module in case the use case implemented was incompatible with the main one. I have no problem changing it.

PR #98, opened 2016-07-14
Return removed

adokando:


pedrobaeza:
Although the return is not needed, it doesn't harm either, so this is not needed. You need to sign the CLA also.

adokando:
This can lead to errors if you put two or more agents and one of them has not formula commission type. For example if we introduce three agents with fixed percentage, formula and fixed percentage commissions, it loops the lines and the first element not being formula type, it returns only the super of first line, in this way it only computes the amount of first line setting zero the amount of the other two agents.

pedrobaeza:
You're right, as the method is implicitly handled as @api.multi. Please sign the CLA then.

adokando:
I already sent the CLA and recived mail from membership@odoo-community.org verifying it. Should I do anything else?

pedrobaeza:
No, sorry, I wasn't aware about that. I fast-track this then.

Thanks for the contribution.

PR #99, opened 2016-08-08
[IMP] sale_commission: updated some views

oihane:


sale_commission/models/sale_order.py
```diff
@@ -84,7 +83,7 @@ def onchange_agent(self):
     @api.depends('sale_line.price_subtotal')
     def _compute_amount(self):
         for line in self:
-            line.amount = 0.0
+            amount = 0.0
```
pedrobaeza: This is not needed, as these values are cached and you don't have performance loss in this case.

sale_commission/views/settlement_view.xml
```diff
@@ -55,35 +56,41 @@
...
+                    <sheet string="Settlement">
+                        <div class="oe_title" >
+                            <h1><field name="agent" /></h1>
+                        </div>
                         <group>
-                            <field name="date_from"/>
-                            <field name="invoice" attrs="{'invisible': [('state', '!=', 'invoiced')]}"/>
+                            <group>
+                                <field name="date_from"/>
+                                <field name="invoice"
```
pedrobaeza: Please don't make cosmetic changes that makes the diff very large

sale_commission/__init__.py
```diff
@@ -4,3 +4,4 @@
 
 from . import models
 from . import wizard
+from .hooks import set_commission_total
```
pedrobaeza: Newline at the end

sale_commission/hooks.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    When there are to many sale.order created it takes a very long time to
+    create the computed field commission_total, in order to ease the
+    installation we create the field manually.
+    :param cr: database cursor
+    :return: void
+    """
+    cr.execute("""
+        ALTER TABLE sale_order
+        ADD COLUMN commission_total double precision;""")
```
pedrobaeza, line 18: You must make the computation also by SQL

oihane, line 18: it depends on fields that will be 0, so commission_total will be 0.

pedrobaeza, line 18: OK, please put a comment about that then

sale_commission/views/settlement_view.xml
```diff
@@ -55,35 +55,40 @@
                                 name="action_cancel" />
                         <field name="state" widget="statusbar"/>
                     </header>
-                    <h1><field name="agent" /></h1>
-                    <group>
+                    <sheet string="Settlement">
```
pedrobaeza, line 58: Why these changes? This was recently changed in https://github.com/OCA/commission/pull/92, so you shouldn't remove it

sale_commission/hooks.py
```diff
@@ -0,0 +1,18 @@
...
+# © 2016 Oihane Crucelaegui - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+
+def set_commission_total(cr):
+    """
+    When there are to many sale.order created it takes a very long time to
+    create the computed field commission_total, in order to ease the
+    installation we create the field manually.
+    As the value of it will be calculated from data that is not created it's
```
pedrobaeza: As its value will be computed from data that still doesn't exist, its value will be always 0, so there's no need to be filled.

sale_commission/views/settlement_view.xml
oihane, line 58: I do not remove it, I just changed the formatting adding a sheet.

pedrobaeza, line 58: OK

oihane:
@pedrobaeza all changes done

pedrobaeza:
:+1:

ivantodorovich:
:+1:  looks good

PR #100, opened 2016-08-26
[8.0][FIX]sale_commission: Only create commissions in customer invoices.

jesusVMayor:
Fixes the creation of account.invoice.line.agent in supplier invoices. It occurs when a customer with agent is also an agent.

### DIFF
--- a/sale_commission_product/models/account_invoice.py
+++ b/sale_commission_product/models/account_invoice.py
@@ -17,8 +17,7 @@ class AccountInvoiceLine(models.Model):
         res = super(AccountInvoiceLine, self).product_id_change(
             product, uom_id, qty, name, type, partner_id,
             fposition_id, price_unit, currency_id, company_id)
-
-        if partner_id and product:
+        if type in ('out_invoice', 'out_refund') and partner_id and product:
             agent_list = []
             partner = self.env["res.partner"].browse(partner_id)
             for agent in partner.agents:
@@ -28,9 +27,10 @@ class AccountInvoiceLine(models.Model):
                     .get_commission_id_product(product, agent)
                 if commission_id_product:
                     commission_id = commission_id_product
-                agent_list.append({'agent': agent.id,
-                                   'commission': commission_id
-                                   })
+                if commission_id:
+                    agent_list.append({'agent': agent.id,
+                                       'commission': commission_id
+                                       })
 
                 res['value']['agents'] = [(0, 0, x) for x in agent_list]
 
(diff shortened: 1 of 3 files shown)


sale_commission/models/account_invoice.py
```diff
@@ -62,26 +62,26 @@ class AccountInvoiceLine(models.Model):
     _inherit = "account.invoice.line"
 
     @api.model
-    def _default_agents(self):
+    def create(self, vals):
```
pedrobaeza: You must not replace the default. It's there for something: modify in screen the default commission values if needed.

sale_commission_product/models/account_invoice.py
```diff
@@ -17,21 +17,22 @@ def product_id_change(
...
-                commission_id_product = self.env["product.product.agent"]\
-                    .get_commission_id_product(product, agent)
-                if commission_id_product:
-                    commission_id = commission_id_product
-                agent_list.append({'agent': agent.id,
-                                   'commission': commission_id
-                                   })
-
-                res['value']['agents'] = [(0, 0, x) for x in agent_list]
+        if type == 'out_invoice':
```
ivantodorovich: what about out_refund?

jesusVMayor:
Added out_refund.
I added a check in the settlement wizard because the only thing that prevents commissions creation for the supplier invoices its the partner_id in contexts it is not passed, and this could be changed by any module. I not see another way to improve the control of commissions creation without removing the default.

sale_commission_product/models/account_invoice.py
```diff
@@ -17,21 +17,22 @@ def product_id_change(
...
-                commission_id_product = self.env["product.product.agent"]\
-                    .get_commission_id_product(product, agent)
-                if commission_id_product:
-                    commission_id = commission_id_product
-                agent_list.append({'agent': agent.id,
-                                   'commission': commission_id
-                                   })
-
-                res['value']['agents'] = [(0, 0, x) for x in agent_list]
+        if type in ('out_invoice', 'out_refund'):
```
pedrobaeza: Don't make 2 nested if, but put all conditions in one.

pedrobaeza (changes requested):

atchuthan (approved):
minor change as per @pedrobaeza suggestion

Could change label to Needs fixing!!

jesusVMayor:
Fixed

pedrobaeza (approved):

PR #104, opened 2016-10-25
[8.0][IMP] Partial Payments

mbcosta:
In some cases of use the company make sells with Payment Terms like this two times in 07 and 28 days or three times in 30, 60, 90 days, the commission was paid or with Invoice created or after all Invoice is Paid, I included other option after the first Partial Payments we can create the Settlement with value based in the same percentage was paid. Example:

Total Invoice  = 20000
Two Payments = 10000 after 7 days and 10000 after 28 days
Total Commission = 4000

After make the first payment
Total of Settlement need to be = 2000
And after second payment
Total of Settlement need to be = 2000

### DIFF
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '8.0.3.0.0',
+    'version': '8.0.4.0.0',
     'author': 'Pexego, '
               'Savoire-faire linux, '
               'Avanzosc, '
@@ -31,7 +31,8 @@
         "Daniel Campos <danielcampos@avanzosc.es>",
         "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
         "Oihane Crucelaegui <oihanecruce@gmail.com>",
-        "Iván Todorovich <ivan.todorovich@gmail.com",
+        "Iván Todorovich <ivan.todorovich@gmail.com"
+        "Akretion - Magno Costa <magno.costa@akretion.com.br>",
     ],
     "data": [
         "security/commission_security.xml",
--- a/sale_commission/__openerp__.py
+++ b/sale_commission/__openerp__.py
@@ -31,7 +31,7 @@
         "Daniel Campos <danielcampos@avanzosc.es>",
         "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
         "Oihane Crucelaegui <oihanecruce@gmail.com>",
-        "Iván Todorovich <ivan.todorovich@gmail.com"
+        "Iván Todorovich <ivan.todorovich@gmail.com>",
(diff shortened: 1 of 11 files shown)


sale_commission/__openerp__.py
```diff
@@ -31,7 +31,8 @@
         "Daniel Campos <danielcampos@avanzosc.es>",
         "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
         "Oihane Crucelaegui <oihanecruce@gmail.com>",
-        "Iván Todorovich <ivan.todorovich@gmail.com",
+        "Iván Todorovich <ivan.todorovich@gmail.com"
```
pedrobaeza: Don't remove comma

atchuthan (approved):
Code review :+1:

sale_commission/models/settlement.py
```diff
@@ -172,8 +179,11 @@ class SettlementLine(models.Model):
         comodel_name="res.partner", readonly=True, related="agent_line.agent",
         store=True)
     settled_amount = fields.Float(
-        related="agent_line.amount", readonly=True, store=True)
+        string='Settled Amount', digits_compute=dp.get_precision('Account'))
```
pedrobaeza, line 182: Why removing this related field?

sale_commission/models/settlement.py
```diff
@@ -14,7 +15,9 @@ class Settlement(models.Model):
     def _default_currency(self):
         return self.env.user.company_id.currency_id.id
 
-    total = fields.Float(compute="_compute_total", readonly=True, store=True)
+    total = fields.Float(
+        compute="_compute_total", readonly=True, store=True,
```
pedrobaeza: A compute field is already readonly without inverse function.

sale_commission/models/settlement.py
```diff
@@ -162,6 +165,10 @@ class SettlementLine(models.Model):
         relation='settlement_agent_line_rel', column1='settlement_id',
         column2='agent_line_id', required=True)
     date = fields.Date(related="agent_line.invoice_date", store=True)
+    effective_date = fields.Date(
+        string="Effective date",
+        help="In case of commission type is based in Payments use this date"
+             " in case of based in Invoice use invoice date.")
```
pedrobaeza: Add dots and commas to the redaction of the help.

pedrobaeza (changes requested):

sale_commission/models/settlement.py
mbcosta: Hi @pedrobaeza, thank you for review, the reason is the related field has the total value to be paid to commission but now with partial payments this field can't be the same because here the value can be 30%, 50% etc depending the prorportional of what was paid in invoice

pedrobaeza:
There are conflicting files and this is very old. I'm going to close it. If you still see this as needed, please fix the conflicts and open a new PR.

PR #108, opened 2016-11-11
change compute period for agent settle

scigghia:
in this way, date range for settle is correct


giudalo:
:+1:

Daniel-CA:
This change is not necessary, the settlement wizard is set to create all settlements until the date defined.
So if your settlements are created monthly if you check last day of the current month is not going to be settled.

pedrobaeza:
You have to put it at least the first day of the next period.

scigghia:
So user have no choice, even if it want to settle only a week for a specific agent. 
is possible that i want to pay in advance one week for a monthly agent or two moth for a quarterly agent .
My opinion is to give freedom to user.

pedrobaeza:
Not this way. That's a change in the behavior. You can add the feature (adding the corresponding selection to be made by the user), but not replacing the current one.

PR #109, opened 2016-12-15
[FIX] wizard_settle create only one settlement per invoice and agent …

enriquemartin:
[FIX] wizard_settle create only one settlement per invoice and agent combination


pedrobaeza:
Closing in favor of #127, which is green in Travis + runbot.

PR #111, opened 2016-12-21
[8.0][FIX] api decorator on module sale_commission

nicomacr:
The method "_refund_cleanup_lines" in odoo is defined like "def _refund_cleanup_lines(self, lines)" and should be decoreted with @api.model instead of @api.multi

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -39,7 +39,7 @@ class AccountInvoice(models.Model):
         settlements.write({'state': 'invoiced'})
         return super(AccountInvoice, self).invoice_validate()
 
-    @api.multi
+    @api.model
     def _refund_cleanup_lines(self, lines):
         """ugly function to map all fields of account.invoice.line
         when creates refund invoice"""


pedrobaeza (approved):

pedrobaeza:
Please check Travis status

jjscarafia (approved):

jjscarafia:
I don't understand why of the errors on the tests, it is not due to this change, could be another module that has change and Interfere on tests?

pedrobaeza:
Well, it's an error on this module, so do this:

* Check first if the error is also on the main branch.
* If not, run the tests locally to debug the problem.

nicomacr:
@pedrobaeza  I don't understand the error. is the same in this PR  https://github.com/OCA/commission/pull/109
, https://github.com/OCA/commission/pull/108

pedrobaeza:
But it's the error also on the main branch?

nicomacr:
By main branch you refer 8.0, don't you?
8.0 also falling due to same error, check here https://travis-ci.org/OCA/commission/branches

pedrobaeza:
Yeah, that's what I mean, so the problem is not due to your change. Maybe you can investigate why it happens and in what commit it started to fail?

Emiq2 (approved):

atchuthan (approved):

pedrobaeza:
OK, let's merge this one, although the branch is still red and anyone should fix it.

PR #116, opened 2017-02-23
[8.0][IMP] Included option to Grouping Invoice by agent

mbcosta:
Today when we click in wizard to create Invoices it create one Invoice to one Settlement, on this PR I included the option on wizard to "Grouping Invoice by Agent" if  the field was checked the program will create one Invoice with all Settlements per Agent.

### DIFF
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -114,6 +114,10 @@ class Settlement(models.Model):
         invoice_line_vals['name'] += "\n" + _('Period: from %s to %s') % (
             date_from.strftime(lang.date_format),
             date_to.strftime(lang.date_format))
+        # invert invoice values if it's a refund
+        if invoice_vals['type'] == 'in_refund':
+            invoice_line_vals['price_unit'] = -invoice_line_vals['price_unit']
+
         return invoice_line_vals
 
     def _add_extra_invoice_lines(self, settlement):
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Can you try to avoid duplicating so many lines of code from the grouping/not grouping options?

mbcosta:
Hi @pedrobaeza , thank you for review. The first code was very ugly, I made some changes to try avoid duplicate lines, please take a look, any sugestions are welcome

pedrobaeza:
I think this shouldn't be done this way and adding this is overkill. If you want to group invoices from different settlements because it's not enough to change settlement period, use `account_invoice_merge` modules instead.

Closing as this is not going to be merged.

mbcosta:
Hi @pedrobaeza , thanks for the reply. If you can help me I would like to know what is better way to implement this, my scenario is:

At end of month the company will pay the commission, the actual code create one invoice per settlement ( sorry if it's wrong, I made the code a long time ago, maybe I missed something ), so if there is 80 settlements to same agent there will be 80 invoices, the users at the company said this is not productive, because they need manage 80 invoices when the ideal is 1 invoice per agent with the total values to be pay and the details about settlements in the lines for conference, and said don't seeing any reason for other behaivor, so I put the option to "grouping invoices per agent" if the checkbox is marked the code will create 1 invoice with 80 lines, and it's more easy for the way they work.  If the option is not marked the code have the same behaivor before the PR.

We have this code in production since the PR was made, and we would want to put this code at OCA, so any ideas to make this work properly are appreciate

pedrobaeza:
As said, this is duplicating the feature of `account_invoice_merge`, and adding all this code here only makes this to be more complicated to maintain.

PR #124, opened 2017-05-31
[REF] [8.0]hr_commission: _get_employee function

maurochip:
Change _get_employee function to use the api multi.


pedrobaeza (approved):
Although this is totally not needed.

pedrobaeza:
Travis is red and as said, it's not needed, so i close this.

PR #125, opened 2017-07-19
[FIX] [8.0] sale_commission: get correct date to do settlements + get corre…

Daniel-CA:
```date_to_agent``` has the starting date so any invoice greater than it will be passed when doing settlements .
To get starting month when quarter period is defined ```*3``` operation is not necesary

### DIFF
--- a/sale_commission/wizard/wizard_settle.py
+++ b/sale_commission/wizard/wizard_settle.py
@@ -22,7 +22,7 @@ class SaleCommissionMakeSettle(models.TransientModel):
             return date(month=date_to.month, year=date_to.year, day=1)
         elif agent.settlement == 'quaterly':
             # Get first month of the date quarter
-            month = ((date_to.month - 1) // 3 + 1) * 3
+            month = (date_to.month - 1) // 3 + 1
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: sale_commission/wizard/wizard_settle.py [hunk 1]
```diff
@@ -63,7 +63,7 @@ def action_settle(self):
             date_to_agent = self._get_period_start(agent, date_to)
             # Get non settled invoices
             agent_lines = agent_line_obj.search(
-                [('invoice_date', '<', date_to_agent),
+                [('invoice_date', '<', date_to),
```
pedrobaeza, line 66: You shouldn't change this because it's on purpose this way.

Daniel-CA: So you need to take next period date to get current ones settlements? Looks a bit weird.

pedrobaeza: No, it isn't. You can't make the settlement of something in progress, as it's still not completed (there are possible invoices to come).

Daniel-CA: Got it, thanks.

PR #133, opened 2017-10-10
[9.0][IMP] sale_commission: Add settlement menu view for sale_manager

oihane:


pedrobaeza (approved):

atchuthan (approved):

pedrobaeza:
Now it only rests 10.0 with this improvement

oihane:
I think it already has it:

https://github.com/OCA/commission/blob/10.0/sale_commission/views/settlement_view.xml#L135

pedrobaeza:
Yeah, right! Then everything it is!

PR #138, opened 2017-12-10
[FIX] Compute taxes

hmrodrigues:
It fixes the bug where invoices are created without total tax


pedrobaeza:
Please check Travis

pedrobaeza:
I'm seeing this patch is also on 10.0, and the 9.0 branch is already broken, so I fast-track it.

PR #142, opened 2018-01-11
[FIX][sale_commission_formula] fix: no encoding declared

ernesto-telematel:


pedrobaeza:
11.0 doesn't require encoding as Python 3 takes already the file encoding. See migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

PR #143, opened 2018-01-22
Ádded translations (es)

EdgarSolorzano:
Corrected and added some translations in spanish.

### DIFF
--- a/sale_commission/i18n/es.po
+++ b/sale_commission/i18n/es.po
@@ -165,6 +165,11 @@ msgid "Commission types"
 msgstr "Tipos de comisión"
 
 #. module: sale_commission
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Please use Transifex instead: https://odoo-community.org/page/translate

EdgarSolorzano:
Hi Pedro, is it possible to fix the "Settlments" (wrong spelled) menu item and the missing Commissions Management keyword through Transifex?

pedrobaeza:
OK, I reopen the PR for only the misspelled word, but no need to refresh POT file (it should be removed in fact).

pedrobaeza:
No answer, so I close.

PR #144, opened 2018-01-23
[11.0][FIX] minor fixes

rgarnau:
This PR fixes:
- In sale_commission prevents the possibility of creating new invoices with commissions belonging to a settled group that has already been invoiced. As, in this case, the generated vendor bill for this group cannot be updated.
- In sale_commission_formula takes into account whether if the invoice has been refunded or not.

@jbeficent @etobella

### DIFF
--- a/sale_commission/tests/test_sale_commission.py
+++ b/sale_commission/tests/test_sale_commission.py
@@ -57,6 +57,12 @@ class TestSaleCommission(TransactionCase):
         self.journal = self.env['account.journal'].search(
             [('type', '=', 'purchase')], limit=1
         )
+        self.agent_monthly = self.res_partner_model.create({
+            'name': 'Test Agent - Monthly',
+            'agent': True,
+            'settlement': 'monthly',
+            'lang': 'en_US',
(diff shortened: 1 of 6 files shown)


JordiBForgeFlow (approved):

JordiBForgeFlow:
@eLBati @hurrinico Can you review? We added full tests

etobella (approved):
LGTM :+1:
Version should be changed

sale_commission/tests/test_sale_commission.py
```diff
@@ -426,3 +426,63 @@ def test_invoice(self):
             agent.agent = self.agent_semi
             agent.onchange_agent()
             self.assertEqual(self.agent_semi.commission, agent.commission)
+
+    def check_new_invoice_with_settle_invoiced(self, agent, commission,
```
pedrobaeza: Where is this method called?

rgarnau:
@etobella version has been changed
@pedrobaeza good catch! already fixed. Thanks!

PR #146, opened 2018-02-12
[ADD] sale_commission_pricelist: New module that assign commission to…

carlosdauden:
… pricelist item

@Tecnativa


sale_commission_pricelist/README.rst
```diff
@@ -0,0 +1,73 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+==============================
+Sales commissions by pricelist
+==============================
+
+This module extends the functionality of ... to support ...
```
pedrobaeza: The description is not correct

sale_commission_pricelist/README.rst
```diff
@@ -0,0 +1,73 @@
...
+You can then make the settlements of these commissions, and generate the
+corresponding supplier invoices to pay their commissions fees.
+
+You can define which base amount is going to be taken into account: net amount
+(based on margin) or gross amount (line subtotal amount)
+
+Usage
+=====
+
+To use this module, you need to:
```
pedrobaeza, line 15: ~~Put usage instructions~~

sale_commission_pricelist/README.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://odoo-community.org/logo.png>`_.
+
+Contributors
+------------
+
+* Carlos Dauden <carlos.dauden@tecnativa.com>
```
pedrobaeza: Put Tecnativa name as first bullet level, and then as second your name.

sale_commission_pricelist/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Sales commissions by pricelist',
+    'version': '10.0.1.0.0',
+    'author': 'Tecnativa, '
+              'Odoo Community Association (OCA)',
+    'category': 'Sales',
```
pedrobaeza: Missing website key

sale_commission_pricelist/models/product_pricelist.py
```diff
@@ -0,0 +1,16 @@
...
+# Copyright 2018 Carlos Dauden - Tecnativa <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProductPricelistItem(models.Model):
+    _inherit = 'product.pricelist.item'
+
+    # commission_id or commission like all references??
```
pedrobaeza: Let's put it with `_id` this time.

sale_commission_pricelist/models/sale_order.py
```diff
@@ -0,0 +1,26 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    @api.onchange('product_id')
+    def _onchange_sale_commission_pricelist(self):
```
pedrobaeza: `_onchange_product_id_sale_commission_pricelist`

sale_commission_pricelist/models/sale_order.py
```diff
@@ -0,0 +1,26 @@
...
+        self.ensure_one()
+        if not self.product_id:
+            return
+        rule_id = self.order_id.pricelist_id.get_product_price_rule(
+            product=self.product_id,
+            quantity=self.product_uom_qty or 1.0,
+            partner=self.order_id.partner_id,
+            date=self.order_id.date_order,
+            uom_id=self.product_uom.id)[1]
```
pedrobaeza, line 21: Why `[1]`? Does this always return that values?

pedrobaeza (changes requested):
Use a different icon for this module.

sale_commission_pricelist/README.rst
carlosdauden: In the PR text you can read "Tests and Readme WIP"
:wink:

sale_commission_pricelist/models/sale_order.py
carlosdauden: It should.
Return a tuple.

pedrobaeza: But it contains always at least 2 elements?

carlosdauden: https://github.com/odoo/odoo/blob/f3da958fa144ee1b112df166435a54c064a10812/addons/product/models/product_pricelist.py#L244

carlosdauden:
Changes done

pedrobaeza (approved):
cc @rafaelbn

sale_commission_pricelist/README.rst
```diff
@@ -0,0 +1,67 @@
...
+
+This module extends the functionality of sale_commission to allow you set a
+commission to pricelist item.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to pricelist
```
rafaelbn: - Go to Sales -> Configuration -> Settings -> Pricing -> Sale Price and activate: "Advanced pricing based on formulas (discounts, margins, rounding)"
- Go to Sales -> Configuration -> Price-list and edit an existing one or create a new one
- In the price-list -> add a new item or open a new one
- You will find the field commission in the price-list item

sale_commission_pricelist/README.rst
```diff
@@ -0,0 +1,67 @@
...
+#. Set commission
+#. Create sale order
```
rafaelbn: I've tested @carlosdauden and I'm not sure if it works:

- I set 20% commision for pricelist but
- Sale order takes 10% commision

carlosdauden: Are you surf that you know steps to use sale_commission?
Try adding Agent to partner :wink:

sale_commission_pricelist/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'website': 'https://github.com/OCA/commission/tree/10.0/'
```
pedrobaeza: Better to put only main repository URL

sale_commission_pricelist/models/sale_order.py
```diff
@@ -0,0 +1,25 @@
...
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = 'sale.order.line'
+
+    @api.onchange('product_id')
+    def _onchange_product_id_sale_commission_pricelist(self):
+        self.ensure_one()
+        if self.product_id:
```
pedrobaeza: Maybe, there's no `pricelist_id` set in your SO, and thus there's an error due to a `self.ensure_one()` in method `get_product_price_rule`. Thus, change this condition to:
```python
if self.product_id and self.order_id.pricelist_id:
```

pedrobaeza (changes requested):

sale_commission_pricelist/models/sale_order.py
carlosdauden: pricelist_id is required
https://github.com/odoo/odoo/blob/4b1ed4493612b04044f64f268e74016cf419182d/addons/sale/models/sale.py#L131

pedrobaeza: Well, I'm saying this because our Gitlab CI has failed with that error, so there should be any condition that doesn't fit. I think it's due to calls to `new` instead of `create`. There, required fields are not applied.

carlosdauden:
@pedrobaeza changes done

pedrobaeza:
What about README texts?

carlosdauden:
This module only adds one field in pricelist item.
Extend readme must be here:
https://github.com/OCA/commission/blob/10.0/sale_commission/README.rst

pedrobaeza:
No, it's the Configuration section. You must put the complete steps. It's not valid to put "Go to pricelist". Where is the pricelist entry. @rafaelbn has already put the steps. You only need to copy/paste them.

carlosdauden:
Changes done.

Should I also explain how to create a partner and a sale order?

pedrobaeza:
Please let the irony mode off. You have to explain things that are key for the module. A partner is not, but the pricelist part is. This is always subjective, of course, but in this case it's clear the separation.

pedrobaeza (approved):

pedrobaeza:
@rafaelbn please finish your review here.

rafaelbn:
Rebuilding for testing

rafaelbn (approved):
Great work thanks! :+1: 

Are we going to improve test @carlosdauden ? codecov/patch — 69.23% of diff hit (target 71.08%)

PR #122, opened 2017-05-05
Porting of module sale commission formula to Version 10.0

hurrinico:
this one depends on

- [x] sale_commission: https://github.com/OCA/commission/pull/117

PR #147, opened 2018-02-15
[MIG] sale_commission_formula: Migration to 10.0

lucode:
There is an issue with the access_token in the demo data, which I couldn't solve.


pedrobaeza:
There's another migration of this module in #122. Please review that one instead.

PR #148, opened 2018-02-27
[IMP] sale_commission: Spanish translation

carlosdauden:
@Tecnativa


pedrobaeza:
This is usually done in Transifex, but I understand that you want to add it in batch reusing your existing one.

pedrobaeza (approved):

PR #149, opened 2018-02-27
[IMP] sale_commission: Improve onchanges

carlosdauden:
@Tecnativa

### DIFF
--- a/sale_commission/README.rst
+++ b/sale_commission/README.rst
@@ -46,6 +46,8 @@ Contributors
 * Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
 * Oihane Crucelaegui <oihanecruce@gmail.com>
 * Nicola Malcontenti <nicola.malcontenti@agilebg.com>
+* `Tecnativa <https://www.tecnativa.com>`_:
+  * Carlos Dauden
 
 Images
 -------
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -3,6 +3,13 @@
 from odoo import api, exceptions, fields, models, _
 
 
+def _execute_onchanges(records, field_name):
+    """Helper methods that executes all onchanges associated to a field."""
+    for onchange in records._onchange_methods.get(field_name, []):
+        for record in records:
+            onchange(record)
+
+
 class Settlement(models.Model):
     _name = "sale.commission.settlement"
     _description = "Settlement"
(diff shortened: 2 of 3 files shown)


sale_commission/models/settlement.py
```diff
@@ -3,6 +3,13 @@
 from odoo import api, exceptions, fields, models, _
 
 
+def _execute_onchanges(records, field_name):
+    """Helper methods that executes all onchanges associated to a field."""
+    for onchange in records._onchange_methods.get(field_name, []):
+        for record in records:
+            onchange(record)
```
yajo, line 10: At times, onchanges return a dict instead of modifying current recordset. That's a supported feature and thus this automation can easily break.

The safest options to do this are:

- Call the method directly: `record._onchange_whatever()`
- Use this technique, which imitates what the web client does:

  ```py
  record.update(record.onchange({}, field_name, {field_name: "1"}))
  ```

pedrobaeza: @Yajo not in new API. Please put me an example.

yajo: Support is implemented in [this method](https://github.com/odoo/odoo/find/11.0#L4875-L4879).

pedrobaeza: I don't see the method

yajo: Yikes, wrong link. See here: https://github.com/odoo/odoo/blob/fcb48b724152db876f6752560cce5a5d875e8f46/odoo/models.py#L4876-L4878

pedrobaeza: OK, but then the code is a bit more complicated. I haven't seen anyway return values in Odoo code.

sale_commission/README.rst
```diff
@@ -46,6 +46,8 @@ Contributors
 * Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
 * Oihane Crucelaegui <oihanecruce@gmail.com>
 * Nicola Malcontenti <nicola.malcontenti@agilebg.com>
+* `Tecnativa <https://www.tecnativa.com>`_:
+  * Carlos Dauden
```
pedrobaeza, line 50: Put an empty line above this line for correct RST syntax and include my name here.

pedrobaeza (changes requested):
Please made the changes commented

pedrobaeza:
I have reduced this PR to the onchanges improvement and merge directly what is undeniable that it's needed - the compute taxes call -  for avoiding conflicts when using all together.

rafaelbn (commented):
I think we should finish this PR as this module is used but lots of Companies

rafaelbn:
Now this PR has conflicts, please @pedrobaeza @carlosdauden merge or close but we should care this in v11 an v12

pedrobaeza:
Let's discard this one as not in use anymore.

PR #150, opened 2018-03-20
[FIX] Sale commission domain error

etobella:
There is a small bug on domain definition.


pedrobaeza (approved):
Please change module version number and I'll fast-track it

etobella:
Version changed

PR #151, opened 2018-04-01
[FIX] sale_commission: Avoid erasing of invoice agents

pedrobaeza:
Some onchanges were introduced in hurrinico#2 for workarounding the issue odoo/odoo#17618, but they are causing that invoices that have been invoiced with the proper commissions propagated from sales order, to empty their values when changing one of the following values:

* Partner
* Journal
* Invoice date
* Payment term

It should be taking into account that there are a lot of extensions to this module that can put different agents according some conditions on sales order, but not on invoices (for example, *sale_commission_formula* with formula only in sales orders, or the new *sale_commission_pricelist*, that bases the commission on the pricelist item used, and that's only applicable on sales orders).

More over, the problem will still appear if you add in other module another onchange.

The solution to properly bypass the issue is to directly not add the field on the invoice form view. This way, there are no glitches. To compensate it, I have add it a computed field that show the same information, or even better, as we reduce the number of fields to 1 instead of 2, and this field can be used for expanding the given information.

### DIFF
--- a/sale_commission/__manifest__.py
+++ b/sale_commission/__manifest__.py
@@ -2,8 +2,11 @@
 
 {
     'name': 'Sales commissions',
-    'version': '10.0.1.1.0',
-    'author': 'Odoo Community Association (OCA)',
+    'version': '10.0.2.0.0',
+    'author': 'Odoo Community Association (OCA),'
+              'Tecnativa,'
+              'AvanzOSC,'
+              'Agile Business Group',
     'category': 'Sales Management',
     'license': 'AGPL-3',
     'depends': [
--- a/sale_commission/i18n/es.po
+++ b/sale_commission/i18n/es.po
@@ -19,6 +19,12 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 #. module: sale_commission
+#: code:addons/sale_commission/models/account_invoice.py:114
+#, python-format
+msgid "%s commission agents"
+msgstr "%s comisionistas"
+
+#. module: sale_commission
 #: model:ir.ui.view,arch_db:sale_commission.sale_commission_make_invoice_form
 msgid "(keep empty for invoicing all the settlements)"
 msgstr "(mantener vacío para facturar todas las liquidaciones)"
@@ -29,9 +35,15 @@ msgid "(keep empty for making the settlement of all agents)"
 msgstr "(mantener vacío para realizar la liquidación de todos los agentes)"
 
 #. module: sale_commission
+#: code:addons/sale_commission/models/account_invoice.py:112
+#, python-format
+msgid "1 commission agent"
+msgstr "1 comisionista"
+
+#. module: sale_commission
 #: model:ir.model.fields,field_description:sale_commission.field_sale_commission_active
 msgid "Active"
-msgstr "Activa"
+msgstr "Activo"
 
 #. module: sale_commission
 #: model:ir.model.fields,field_description:sale_commission.field_account_invoice_line_agent_agent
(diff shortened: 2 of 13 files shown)


pedrobaeza:
Ouch, I'm now seeing the problem removing these onchanges. I'll study an alternative.

pedrobaeza:
OK, now this is working correctly simply removing the field `agents` from invoice view. Please see main comment for a complete description and the field replacement.

pedrobaeza:
I will change the version number on merge for avoiding conflicts with other opened PRs.

pedrobaeza:
Seeing that the same problem with onchanges can happen on sales orders, it's better to join all the common logic in sales and invoices in some mixins for not duplicating code, and apply the same on both.

Some fields are still specific to the sales order or invoice objects, as same as the views, but it's a good code reduction.

I have made this refactoring on a separate commit.

lmignon:
If I remember well, we were faced this problem in product_variant_configurator.

pedrobaeza:
Yes, the pattern is the same: it has a nested one2many into another one2many. With this workaround, it doesn't crash, but it loses some data if you don't save meanwhile.

angelmoya (approved):
Functionally tested on runbot

pedrobaeza:
Does the rest of the proposed reviewers accept to merge this for avoiding the huge problem? (that is also present in v9 and v11)

eLBati:
Hi @pedrobaeza , thanks for your work.

Trying at http://3326892-151-995364.runbot2.odoo-community.org the following steps:
 - Configured customer 'Agrolait' with 2 Agents
 - Configured customer 'Camptocamp' with a third agent
 - Create a sale order, select Agrolait and add a product
 - Without saving the SO, change customer to 'Camptocamp' and save

Get
```
The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set
[object with reference: commission - commission]
```

Do you think it is possible to mitigate?

Thanks

pedrobaeza:
@eLBati I have fixed also the case you expose in https://github.com/OCA/commission/pull/151#issuecomment-391948734. Can we finally merge this?

sale_commission/__manifest__.py
```diff
@@ -2,8 +2,11 @@
 
 {
     'name': 'Sales commissions',
-    'version': '10.0.1.1.0',
-    'author': 'Odoo Community Association (OCA)',
+    'version': '10.0.2.0.0',
+    'author': 'Odoo Community Association (OCA),'
+              'Tecnativa,'
+              'AvanzOSC,'
+              'Agile BG',
```
eLBati: Thanks for the credit. Please use `Agile Business Group` if not too long.

sale_commission/i18n/es.po
```diff
@@ -18,6 +18,12 @@ msgstr ""
 "Content-Transfer-Encoding: \n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
+#. module: sale_commission
+#: code:addons/sale_commission/models/sale_commission_mixin.py:55
+#, python-format
+msgid "%s commission agents"
+msgstr "%s comisionistas"
```
eLBati, line 25: Changes to .po files should be only done by weblate

eLBati (changes requested):
@pedrobaeza thanks. According to my tests, this is OK.
Do you want to keep the 4 commits?

eLBati:
@pedrobaeza also: v11 is affected too. Please consider porting the fix

sale_commission/i18n/es.po
pedrobaeza: Not necessarily. If the PR is properly rebased, there won't be conflict.

eLBati: @sbidoul told me so :innocent: 

Please tell us the rule defining when we can change .po files directly and when we can not.
cc @primes2h @SimoRubi

pedrobaeza: Trust me on this one :wink: The problem might appear if you work on an outdated base and merging Weblate translations with these ones a conflict arises, but that's detected also by Travis.

sale_commission/__manifest__.py
pedrobaeza: Yeah, you have worked a lot over these modules, so I think you deserve it.

pedrobaeza:
@eLBati your author name has been changed. In this case, I prefer to leave the commit history as is, as it belong to different refactoring works.

I'm now going to check v11, as it's the target for my original customer ticket that has made me to review all of this. I was expecting that this doesn't happen in v11 due to the new JS views :cry:

pedrobaeza:
Can we merge then?

eLBati (approved):
Thanks

pedrobaeza:
@eLBati I'm just trying on v11, removing the hacks, and there's no such problem as I was expecting, so that are good news. Any way, I'm testing deeper the module, and adding some refactoring and ideas done in v12 migration.

PR #160, opened 2018-05-21
[8.0] fix multicompany agents search

sergiocorato:


pedrobaeza:
Multi-company record rule is best suited for achieving this than this filter.

sergiocorato:
Ok thanks

PR #161, opened 2018-05-21
[10.0] fix multicompany agents search

sergiocorato:


pedrobaeza:
The same comment as in 8.0 version

sergiocorato:
Ok, thanks

PR #168, opened 2018-07-05
[FIX] hr_commission: Avoid AccessError on partner form

yajo:
If user has no access to employees, the field should be empty, but he should still be able to user ``res.partner`` form and everything else.

@Tecnativa

### DIFF
--- a/hr_commission/models/res_partner.py
+++ b/hr_commission/models/res_partner.py
@@ -30,17 +30,17 @@ class ResPartner(models.Model):
         selection_add=[("salesman", "Salesman (employee)")])
     employee = fields.Many2one(
         comodel_name="hr.employee", compute="_get_employee")
-    users = fields.One2many(comodel_name="res.users",
-                            inverse_name="partner_id")
 
     @api.one
-    @api.depends('users')
+    @api.depends('user_ids')
     def _get_employee(self):
-        self.employee = False
-        if len(self.users) == 1 and len(self.users[0].employee_ids) == 1:
-            self.employee = self.users[0].employee_ids[0]
+        try:
+            self.employee = self.user_ids[:1].employee_ids[:1]
+        except exceptions.AccessError:
+            # You have no access to employees; then it's empty
+            pass
 
-    @api.constrains('agent_type', 'users')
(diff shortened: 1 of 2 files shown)


hr_commission/models/res_partner.py
```diff
@@ -30,17 +30,17 @@ class ResPartner(models.Model):
         selection_add=[("salesman", "Salesman (employee)")])
     employee = fields.Many2one(
         comodel_name="hr.employee", compute="_get_employee")
-    users = fields.One2many(comodel_name="res.users",
```
pedrobaeza, line 33: Why removing this? `user_ids` doesn't contain the same value as this one

yajo: I don't see the difference, which is it?

https://github.com/OCA/OCB/blob/a23f87326a89849c23531e933488ec760ee27e7c/openerp/addons/base/res/res_partner.py#L305

pedrobaeza: You're right. I got confused. Any way, this should be more focused on recent versions, as we are migrating to them.

yajo: Well, it's done already, I don't think it hurts to clean this version... :thinking: 

It should be forward-ported, of course.

hr_commission/models/res_partner.py
```diff
@@ -30,17 +30,17 @@ class ResPartner(models.Model):
...
+        try:
+            self.employee = self.user_ids[:1].employee_ids[:1]
+        except exceptions.AccessError:
+            # You have no access to employees; then it's empty
+            pass
 
-    @api.constrains('agent_type', 'users')
+    @api.constrains('agent_type', 'user_ids')
     def _check_employee(self):
         if self.agent_type == 'salesman' and not self.employee:
```
pedrobaeza, line 45: This is going to give an error if you don't have access to employees then, so it's the same problem

yajo: Then what can I do? Do I use `compute_sudo` instead?

pedrobaeza:
Closing as not spotting this version anymore.

PR #169, opened 2018-07-12
[10.0] FIX sale_commission the following use case

eLBati:
- Go to Sales -> Customers
 - Filters -> Add Custom Filter: Agents contains test

Get

odoo/addons/base/res/res_partner.py", line 647, in name_search
    where_query = self._where_calc(args)
odoo/models.py", line 4038, in _where_calc
    domain = [('active', '=', 1)] + domain
TypeError: can only concatenate list (not "str") to list

### DIFF
--- a/sale_commission/__manifest__.py
+++ b/sale_commission/__manifest__.py
@@ -2,7 +2,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '10.0.1.0.7',
+    'version': '10.0.1.1.0',
     'author': 'Odoo Community Association (OCA)',
     'category': 'Sales Management',
     'license': 'AGPL-3',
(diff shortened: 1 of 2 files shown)


tafaRU (approved):

atchuthan (approved):

eLBati:
Making travis green with https://github.com/OCA/commission/pull/171

tafaRU:
@eLBati, could you please increase the version number?

eLBati:
@tafaRU done

SimoRubi (approved):
Actually this is a core bug, [here](https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L2284) it says that it can be a string too, so someone should report it!
Anyway, here you definitely fix that, so ok for me

pedrobaeza:
Yeah, this is fixing the symptom, not the cause, but in this case we can change it. Domains in strings are put when they have to be evaluated on the fly, not statically.

PR #174, opened 2018-08-06
[MIG][11.0] hr_commission: Migration to 11.0

pedrobaeza:
@Tecnativa


hr_commission/models/res_partner.py
```diff
@@ -1,55 +1,35 @@
...
-            self.employee = self.users[0].employee_ids[0]
+    @api.depends('user_ids')
+    def _compute_employee_id(self):
+        for partner in self:
+            if (len(partner.user_ids) == 1 and
+                    len(partner.user_ids[0].employee_ids) == 1):
+                partner.employee_id = partner.user_ids[0].employee_ids[0]
 
-    @api.constrains('agent_type', 'employee')
+    @api.constrains('agent_type')
```
lmignon: @pedrobaeza All field used by the constrains must be specified into `@api.constrains` to trigger the constrains check if value change. `employee_id` must be in `@api.constrains

lmignon (changes requested):

hr_commission/models/res_partner.py
pedrobaeza: As that field is a computed non-stored value, it shouldn't be there, and a warning reminds you about that. This is how I saw it.

lmignon: Therefore you should add the field declare into the depends of the compute method...

pedrobaeza: But the compute is precisely for computing that field...

lmignon: The code should look like:

```python
   @api.depends('user_ids')
    def _compute_employee_id(self):
        for partner in self:
            if (len(partner.user_ids) == 1 and
                    len(partner.user_ids.employee_ids) == 1):
                partner.employee_id = partner.user_ids.employee_ids

    @api.constrains('agent_type', 'user_ids)
    def _check_employee(self):
        if self.agent_type == 'salesman' and not self.employee_id:
            raise exceptions.ValidationError(
                _("To be salesman there must be one (and only one) employee linked to this "
                  "partner. To do this, go to 'Human Resources' and check "
                  "'Employees'"))
```

If `user_ids` is not into `api.constrains` and you remove/modify the link between `res.parner` and `hr.employee` the constrains will not be triggered.

pedrobaeza: Thanks for pointer. Now I know what you are talking about, but are you sure adding the one2many field to the constraint, it will be triggered on that case? I can develop a test case any way.

lmignon: >  but are you sure adding the one2many field to the constraint, it will be triggered on that case? I can develop a test case any way.

Not sure but a test will be welcome.

pedrobaeza: I have tried and it's not triggered, so I'm not sure about adding a lot of code in employee part just for that corner case. Do you see it so important to do it?

lmignon: IMO, yes it's important to avoid know bug.

pedrobaeza:
@lmignon I have added the code in hr.employee model for checking the condition and test the use case.

pedrobaeza:
@lmignon can you finish your review?

HaraldPanten (approved):
I would add some extra description, maybe. But functionally is working fine :+1:

cristinamartinrod (approved):
Works fine!

lmignon (approved):

PR #175, opened 2018-08-07
[11.0][IMP] Adding graph view and settlement report

aitorbouzas:
Created graph and pivot view for settlements extracted from invoices. Added also a qweb-pdf report for the settlements.

@pedrobaeza there you go :)


aitorbouzas:
@pedrobaeza how can I pass through SQL injection problem? Please check my code, the problem is that psycopg2 `execute(sql, params)` encloses the param values into single quotes...

pedrobaeza:
Hi @aitorbouzas that's not the thing that makes Travis to fail, but any way, you can solve it this way:
```python
from psycopg2.extensions import AsIs
...
query = "CREATE or REPLACE VIEW %%s as ( %s FROM ( %s ) %s );" % (
    self._select(), self._from(), self._group_by(),
)
self._cr.execute(query, (AsIs(self._table), ))
```
The real problem in LINT stage is:
```
sale_commission/report/sale_commission_analysis_report.py:99:80: E501 line too long (94 > 79 characters)
```
And concerning test stages, with this patch, you'll get a proper SQL stage, so it should be OK.

aitorbouzas:
Thank you @pedrobaeza.
Just solved it.

P.S: I already signed the CLA..

sale_commission/report/sale_commission_analysis_report.py
```diff
@@ -0,0 +1,102 @@
...
+from psycopg2.extensions import AsIs
+
+
+class SaleCommissionAnalysisReport(models.Model):
+    _name = "sale.commission.analysis.report"
+    _description = "Sale Commission Analysis Report"
+    _auto = False
+    _rec_name = 'commission_id'
+
+    invoice_state = fields.Selection([
```
pedrobaeza: For not hardcoding this and not being tolerant to extensions, what you can do it put a `selection` method and use the same technique as here:

https://github.com/OCA/sale-workflow/blob/96956205fcd1474443f93da21aaff300f7bb0282/sale_order_type/models/sale_order_type.py#L16-L18

sale_commission/report/sale_commission_analysis_report.py
```diff
@@ -0,0 +1,102 @@
...
+            aila.settled,
+            aila.commission commission_id
+        """
+        return select_str
+
+    def _from(self):
+        from_str = """
+            account_invoice_line_agent aila
+            LEFT JOIN account_invoice_line ail ON ail.id = aila.invoice_line
+            LEFT JOIN account_invoice ai ON ai.id = ail.invoice_id
```
pedrobaeza: This one can be `INNER JOIN`

sale_commission/report/sale_commission_analysis_report.py
```diff
@@ -0,0 +1,102 @@
...
+            LEFT JOIN sale_commission sc ON sc.id = aila.commission
+            LEFT JOIN product_product pp ON pp.id = ail.product_id
+            LEFT JOIN product_template pt ON pp.product_tmpl_id = pt.id
```
pedrobaeza: As well as this one

sale_commission/report/sale_commission_analysis_report.py
```diff
@@ -0,0 +1,102 @@
...
+            ai.state,
+            ai.date_invoice,
+            ail.company_id,
+            rp.id,
+            pt.categ_id,
+            ail.product_id,
+            pt.uom_id,
+            ail.quantity,
+            sc.fix_qty,
+            aila.amount,
```
pedrobaeza: Why do you put this amount in `group by`? All measures shouldn't be in group by.

pedrobaeza (changes requested):
Is this analyzing settlements or invoices generated from the settlements? I think it should be the first one, with additional filter for seeing what is already invoiced.

aitorbouzas:
@pedrobaeza actually this pivot let's you analyze line invoices that has agent commissions in them, and you can check whether it was settled or not. The initial need was to be able to estimate the amount of commissions you must pay in the future and how much did you pay.
Want me to make it so you can analyze settlements?

aitorbouzas:
Another thing: is it necessary to allow users the creation of settlements manually? I could edit the list view so we don't allow it.

pedrobaeza:
> this pivot let's you analyze line invoices that has agent commissions in them

OK, I didn't get the meaning of this. Yeah, I think it's a good idea having the filter to see if they are settled or not.

> Another thing: is it necessary to allow users the creation of settlements manually? I could edit the list view so we don't allow it.

Yes, we should prevent it as they are populated through the wizard.

aitorbouzas:
@pedrobaeza Made all the requested changes, let's see Travis and Runbot. Also edited tree view to restrict manual creation of settlements.

aitorbouzas:
Green! How's it now?

pedrobaeza:
Integrated in #190

PR #176, opened 2018-09-03
[IMP] Add two hooks on make settlements

etobella:
It only adds two hooks on the creation of settlements.

### DIFF
--- a/sale_commission/wizard/wizard_settle.py
+++ b/sale_commission/wizard/wizard_settle.py
@@ -48,7 +48,7 @@ class SaleCommissionMakeSettle(models.TransientModel):
             raise exceptions.Warning(_("Settlement period not valid."))
 
     def _get_settlement(self, agent, company, sett_from, sett_to):
-        return self.env['account.invoice.line.agent'].search([
+        return self.env['sale.commission.settlement'].search([
             ('agent', '=', agent.id),
             ('date_from', '=', sett_from),
             ('date_to', '=', sett_to),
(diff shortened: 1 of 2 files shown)


sale_commission/wizard/wizard_settle.py
```diff
@@ -47,6 +47,23 @@ def _get_next_period_date(self, agent, current_date):
...
+    def _get_settlement_domain(self, agent, company, sett_from, sett_to):
+        return [
+            ('agent', '=', agent.id),
+            ('date_from', '=', sett_from),
+            ('date_to', '=', sett_to),
+            ('company_id', '=', company.id),
+            ('state', '=', 'settled')
+        ]
+
+    def _get_settlement_values(self, agent, company, sett_from, sett_to):
```
pedrobaeza: Call it for following the same practice as Odoo core, `_prepare_settlement_vals`

sale_commission/wizard/wizard_settle.py
```diff
@@ -47,6 +47,23 @@ def _get_next_period_date(self, agent, current_date):
         else:
             raise exceptions.Warning(_("Settlement period not valid."))
 
+    def _get_settlement_domain(self, agent, company, sett_from, sett_to):
```
pedrobaeza: Instead of the domain, return the record you want, as you might change order or other requirement for getting a different settlement.

pedrobaeza (changes requested):

etobella:
@pedrobaeza Changes have been applied.

JordiBForgeFlow (approved):

pedrobaeza (approved):

PR #179, opened 2018-10-05
[12.0][MIG] sale_commission: Migration to 12.0

### LINKED ISSUE #24702: [11.0][web] Incorrect treatment of x2many field that exists in the tree editable view, but not in the form view.
Affects to 11.0

When you disable the 'editable' attribute of the tree associated to the field 'invoice_line_ids' in https://github.com/odoo/odoo/blob/11.0/addons/account/views/account_invoice_view.xml#L412
in a separate module, such as in https://github.com/OCA/commission/blob/11.0/sale_commission/views/account_invoice_view.xml#L44.

When you try to create an invoice and add a new line you'll get the following error message:

image

The root cause is that in this line:
https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/js/views/basic/basic_model.js#L3573
The js expects to retrieve info from the field 'analytic_tag_ids', which currently does not exist in the form view, but it does in the tree view.

The fields to consider for fetching are obtained here:
https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/js/views/basic/basic_model.js#L219

This error occurs for any x2many field that exists in the tree editable view, but not in the form view.

aitorbouzas:
Initial migration.

This is fixed: [https://github.com/odoo/odoo/issues/24702](https://github.com/odoo/odoo/issues/24702) now

Also will add pivot view for settlements as in version 11.0 (PR is in reviewing process at #175 ) when the #175 gets merged.


aitorbouzas:
@pedrobaeza Added the button we talked about, did you mean something like that?

pedrobaeza:
This should be re-done from #190. I will understand if you now decline to make the work, so just tell me and I'll take over.

aitorbouzas:
I must decline the work since i'm totally out of time lately. If I can i'll work in the other PR's i've got.

pedrobaeza:
Superseeded by #193 with all the changes from v11.

PR #181, opened 2018-10-29
[10.0] IMP sale_commission: allow to find partners without (or with) assigned agents

eLBati:


eLBati:
For 11.0: https://github.com/OCA/commission/pull/182

pedrobaeza:
Well, this is very simple with custom search on "Agents" field to "Set" or "Not Set". Having a lot of filters in partner makes this very unusable, so I don't consider this is very needed for all days operations.

eLBati:
image

@pedrobaeza not really a lot of filters IMO.
If this search is done frequently, the filter is convenient, instead of making a custom filter.
User could `save current search`, but it is uncomfortable to combine with other filters

pedrobaeza:
This is because there are no other modules installed in that instance that adds the same filters as yours. I'm not talking about saving filter, but going to "Add Custom Filter" and filter by that field. I'm still against this change.

eLBati:
> This is because there are no other modules installed in that instance that adds the same filters as yours

With many modules, would the dropdown filter menu go out of page frame?

> going to "Add Custom Filter" and filter by that field

I understood. This is even longer :sweat_smile: 

Anyway, no problem. We can put it in a custom module.

eLBati:
Closing and including in https://github.com/OCA/commission/pull/183

PR #188, opened 2018-11-13
[FIX]sale_commission: When creating agent from agents fields, on part…

javierjcf:
…ner view, default field agent must be true


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this as obvious.

PR #189, opened 2018-11-20
[IMP] Add hook on the create settlements

etobella:
This PR adds a hook in order to create other systems.

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -221,3 +221,15 @@ class AccountInvoiceLineAgent(models.Model):
         ('unique_agent', 'UNIQUE(invoice_line, agent)',
          'You can only add one time each agent.')
     ]
+
+    @api.multi
+    def _skip_settlement(self):
+        """
+        This function should return if the commission can be payed
+        :return: bool
+        """
+        self.ensure_one()
+        return bool(
+            self.commission.invoice_state == 'paid' and
+            self.invoice.state != 'paid'
+        )
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -229,7 +229,7 @@ class AccountInvoiceLineAgent(models.Model):
         :return: bool
         """
         self.ensure_one()
-        return bool(
+        return (
             self.commission.invoice_state == 'paid' and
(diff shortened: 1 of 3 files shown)


sale_commission/models/account_invoice.py
```diff
@@ -221,3 +221,15 @@ def _compute_company(self):
...
     ]
+
+    @api.multi
+    def _skip_settlement(self):
+        """
+        This function should return if the commission can be payed
+        :return: bool
+        """
+        self.ensure_one()
+        return bool(
```
pedrobaeza: No need to cast to bool. It's already a boolean.

pedrobaeza (changes requested):
We have agent types for discriminating the payment of settlements, but I don't get why do you want to skip settlements at all.

etobella:
It could be used for the following: 

- Pay agents when the invoice is payed or when the due_date has been accomplished.
- Use the same formula for all agents (in order to avoid duplicity) but you want that some agents are payed as soon as it is invoiced.
- You want to create a different way to manage when to pay the invoices

pedrobaeza:
> Pay agents when the invoice is payed or when the due_date has been accomplished.

OK, good use case, but I think this one should be included in the base module (that's to say, adding conditions in the method you have created). Don't you think?

> Use the same formula for all agents (in order to avoid duplicity) but you want that some agents are payed as soon as it is invoiced.

OK, the same about integrating this through an option at agent level.

> You want to create a different way to manage when to pay the invoices

I don't understand this one, as this is not for paying, but for settling the commissions.

etobella:
The problem is that the system does not block settlements. When you create the settlement, it will be invoiced as soon as you launch the wizard.
So, if you want to block some `account.invoice.line.agent` to be payed, you should not allow to create the settlement. If your company has some conditions in order to decide when to create them, you use this hook in order to define it.

pedrobaeza:
Well, you can do the following:

- Invoice the settlements you want selecting them and clicking on the action.
- Add a new settlement state/check for blocking them.

But now I understand what you want to achieve in that third point.

pedrobaeza (approved):

JordiBForgeFlow (approved):

PR #192, opened 2018-12-07
[10.0][FIX] sale_commission: Journal check

SimoRubi:
Automated tests of 10.0 are failing:
```
ERROR: test_sale_commission_net_amount_invoice (odoo.addons.sale_commission.tests.test_sale_commission.TestSaleCommission)
Traceback (most recent call last):
`   File "/home/travis/build/OCA/commission/sale_commission/tests/test_sale_commission.py", line 286, in test_sale_commission_net_amount_invoice
`     wizard2.button_create()
`   File "/home/travis/build/OCA/commission/sale_commission/wizard/wizard_invoice.py", line 46, in button_create
`     self.journal, self.product, date=self.date)
`   File "/home/travis/build/OCA/commission/sale_commission/models/settlement.py", line 127, in make_invoices
`     settlement, invoice_journal, date=date)
`   File "/home/travis/build/OCA/commission/sale_commission/models/settlement.py", line 71, in _prepare_invoice_header
`     'type': ('in_invoice' if journal.type == 'purchase' else
` AttributeError: 'bool' object has no attribute 'type'
```
This is because the journal suggested by the wizard is discarded in https://github.com/OCA/commission/blob/10.0/sale_commission/models/settlement.py#L123.

The `settlement.total + extra_total` was negative because `invoice.line.product_id.standard_price` was being used instead of `invoice.line.price_unit`

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -121,22 +121,15 @@ class AccountInvoiceLineAgent(models.Model):
     @api.depends('object_id.price_subtotal')
     def _compute_amount(self):
         for line in self:
-            line.amount = 0.0
-            if (not line.object_id.commission_free and
-                    line.commission):
-                if line.commission.amount_base_type == 'net_amount':
-                    subtotal = (line.object_id.price_subtotal -
-                                (line.product.standard_price *
-                                 line.object_id.quantity))
-                else:
-                    subtotal = line.object_id.price_subtotal
-                if line.commission.commission_type == 'fixed':
-                    line.amount = subtotal * (line.commission.fix_qty / 100.0)
-                else:
-                    line.amount = line.commission.calculate_section(subtotal)
-                # Refunds commissions are negative
-                if line.invoice.type in ('out_refund', 'in_refund'):
-                    line.amount = -line.amount
+            line.amount = self._get_commission_amount(
+                line.commission,
+                line.object_id.price_subtotal,
+                line.object_id.commission_free,
+                line.product,
+                line.object_id.quantity)
+            # Refunds commissions are negative
+            if line.invoice.type in ('out_refund', 'in_refund'):
+                line.amount = -line.amount
 
     @api.depends('agent_line', 'agent_line.settlement.state', 'invoice',
                  'invoice.state')
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from odoo import api, exceptions, fields, models, _
+from odoo.exceptions import UserError
 
 
 class Settlement(models.Model):
@@ -123,6 +124,10 @@ class Settlement(models.Model):
             invoice_journal = (journal if
                                (settlement.total + extra_total) >= 0 else
                                False)
+            if not invoice_journal:
+                raise UserError(
+                    _('Journal %s is not applicable for quantity %s')
+                    % (journal.display_name, settlement.total + extra_total))
             invoice_vals = self._prepare_invoice_header(
                 settlement, invoice_journal, date=date)
             invoice = invoice_obj.create(invoice_vals)
--- a/sale_commission/tests/test_sale_commission.py
+++ b/sale_commission/tests/test_sale_commission.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from odoo import exceptions, fields
-import openerp.tests.common as common
+import odoo.tests.common as common
 import dateutil.relativedelta
 
 
@@ -285,6 +285,7 @@ class TestSaleCommission(common.TransactionCase):
         wizard2 = self.make_inv_model.create({'product': 1})
         wizard2.button_create()
         settlements = self.settle_model.search([('state', '=', 'invoiced')])
+        self.assertTrue(settlements)
         for settlement in settlements:
             self.assertNotEquals(len(settlement.invoice), 0,
                                  "Settlements need to be in Invoiced State.")
--- a/sale_commission/__manifest__.py
+++ b/sale_commission/__manifest__.py
@@ -2,7 +2,7 @@
 
 {
     'name': 'Sales commissions',
-    'version': '10.0.2.0.0',
+    'version': '10.0.2.1.0',
     'author': 'Odoo Community Association (OCA),'
               'Tecnativa,'
               'AvanzOSC,'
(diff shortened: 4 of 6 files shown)


sale_commission/models/account_invoice.py
```diff
@@ -118,15 +118,19 @@ class AccountInvoiceLineAgent(models.Model):
         compute="_compute_settled",
         store=True, copy=False)
 
-    @api.depends('object_id.price_subtotal')
+    @api.depends('commission', 'commission.amount_base_type',
+                 'commission.commission_type', 'commission.fix_qty',
+                 'invoice.type', 'object_id.commission_free',
+                 'object_id.price_subtotal', 'object_id.price_unit',
+                 'object_id.quantity')
```
pedrobaeza: Why do we need here `object_id.quantity`?

In other hand, changing commissions type shouldn't modify retroactively all the commissions amount already computed, so you should remove that from the depends

pedrobaeza (changes requested):

sale_commission/models/account_invoice.py
SimoRubi: Thanks @pedrobaeza for reviewing.
Here I added all the fields that are involved in the computation of `amount` (that is, all the field involved in this method) because changing any of them might change the value of `amount`.
Changing `object_id.quantity` might change computed value of `amount` so it should trigger a recomputation.

If the recomputation of `amount` shouldn't be done under some circumstances, those circumstances should be checked inside the method to prevent the recomputation.

pedrobaeza: > Changing object_id.quantity might change computed value of amount so it should trigger a recomputation.

If quantity changes the amount, then the trigger will be risen any way, so don't add extra burden to the compute method.

> If the recomputation of amount shouldn't be done under some circumstances, those circumstances should be checked inside the method to prevent the recomputation.

No, you shouldn't check anything inside the method. You simply shouldn't trigger the compute method for that data. As said, you don't retroactively to recompute every past amount when you change the commission rate or similar. That's simply out of business logic.

SimoRubi: > the trigger will be risen any way

How will it be risen if the method `_compute_amount` does not _depend_ on `object_id.quantity`?

> don't retroactively to recompute every past amount when you change the commission rate or similar

I agree, so maybe inside the `_compute_amount` we should execute the computation of `amount` only for the lines (`account.invoice.line.agent` records) that have not _already been computed_, maybe checking the `settled` field?

pedrobaeza: > How will it be risen if the method _compute_amount does not depend on object_id.quantity?

Because it depends on `object_id.price_subtotal`, so it's a transitive effect:

object_id.quantity > object_id.price_subtotal > amount.

> I agree, so maybe inside the _compute_amount we should execute the computation of amount only for the lines (account.invoice.line.agent records) that have not already been computed, maybe checking the settled field?

But that's an specific use case, but I don't want the change to be applied even if they are not settled in my use cases. This should be applied from now on when you do the change. If you want to recompute it, do it with an script, or add a tool for recomputing amounts. And not only this:

- User will experiment a freeze on Odoo saving this value without knowing what is happening.
- This change is too much behavior change for an stable version (but I repeat, I won't do it even on new version due to the stated reasons).

SimoRubi: ok @pedrobaeza, I restored the previous behavior in https://github.com/OCA/commission/pull/192/commits/ed661de4e59aa18c6a0d1b3ed9a7218a385120b9

sale_commission/models/settlement.py
```diff
@@ -122,7 +123,11 @@ def make_invoices(self, journal, product, date=False):
             extra_total = sum(x['price_unit'] for x in extra_invoice_lines)
             invoice_journal = (journal if
                                (settlement.total + extra_total) >= 0 else
-                               False)
+                               journal.browse())
```
pedrobaeza: No need to change this

sale_commission/tests/test_sale_commission.py
```diff
@@ -285,6 +285,7 @@ def test_sale_commission_net_amount_invoice(self):
         wizard2 = self.make_inv_model.create({'product': 1})
         wizard2.button_create()
         settlements = self.settle_model.search([('state', '=', 'invoiced')])
+        self.assertGreater(len(settlements), 0)
```
pedrobaeza: You can do: `self.assertTrue(settlements)`

pedrobaeza (changes requested):

sale_commission/models/account_invoice.py
```diff
@@ -126,7 +126,7 @@ def _compute_amount(self):
                     line.commission):
                 if line.commission.amount_base_type == 'net_amount':
                     subtotal = (line.object_id.price_subtotal -
-                                (line.product.standard_price *
+                                (line.object_id.price_unit *
```
pedrobaeza: I'm checking this right now and the idea behind this is to get the net amount, so it's correct to put the cost price, not the price unit of the line.

SimoRubi: @pedrobaeza thanks for checking, then why are the tests failing?
As said in the description, `settlement.total + extra_total` is negative, then the `journal` is discarded.
Now the code raises an exception for that, then in the test I can use `assertRaises`, but then we will need a test for the complete flow

sale_commission/models/settlement.py
SimoRubi: Done

sale_commission/models/account_invoice.py
pedrobaeza: Well, this is because cost price must be lower than price_unit. By law, you can't sell under the cost price, but any way, we should cover the case for possible wrong data (or incorrect operations but done by other reasons - for example, the total of the operation). For me, in that case, we should make `0` the commission amount.

sale_commission/tests/test_sale_commission.py
SimoRubi: Much cleaner, done

sale_commission/models/account_invoice.py
SimoRubi: @pedrobaeza I implemented your suggestion in https://github.com/OCA/commission/pull/192/commits/0fca4c3fad2d98aa96fb887c75cb46de9c9a7091.

I also moved the commission computation to `sale.commission.line.mixin` because I noticed that it was almost duplicated in `account.invoice` and `sale.order` (almost because `line.object_id` were two different models), now what is happening in the module looks a bit more clear to me; what do you think?
Thanks

pedrobaeza (commented):
I already unified the amount computation in v11, but I like more your approach. I'll forward-port it to v11 and use it on the v12 migration that I'm doing right now. The only problem I see is about the name you have used for the method: `_compute_*` is for computed methods. In this case, it's a method that returns a value, so better to call it `_get_commission_amount`

SimoRubi:
@pedrobaeza, name of the method changed

pedrobaeza (approved):

pedrobaeza:
Thanks for your patience! Merging. You can now check v12 migration in #178. There are several interesting new features like the UI.

PR #193, opened 2018-12-14
[12.0][MIG] sale_commission: Migration to 12.0

pedrobaeza:
* Standard procedure
* Add description to all models
* Adapt commission amount computation courtesy of Simone Rubi
* Adapt views
* Adapt code for UoM and dates
* Fixed dependencies in constraint

cc @Tecnativa


pedrobaeza:
@eLBati @aitorbouzas can you review this? or any opposition to merge this optimistically?

sale_commission/models/settlement.py
```diff
@@ -181,7 +182,7 @@ class SettlementLine(models.Model):
         related='settlement.company_id',
```
eLBati, line 182: @pedrobaeza in v11 this has `readonly=True`. Isn't it needed here?

pedrobaeza: No, this is the default in v12, and thus, not needed.

sale_commission/report/sale_commission_analysis_report.py
```diff
@@ -43,22 +43,23 @@ def _get_selection_invoice_state(self):
...
+            ail.product_id AS product_id,
+            pt.uom_id AS uom_id,
+            SUM(ail.quantity) AS quantity,
+            AVG(ail.price_unit) AS price_unit,
+            SUM(ail.price_subtotal) AS price_subtotal,
+            AVG(sc.fix_qty) AS percentage,
+            SUM(aila.amount) AS amount,
+            ail.id AS invoice_line_id,
+            aila.settled AS settled,
+            aila.commission AS commission_id
```
eLBati, line 62: This means v11 is bugged too, right?

pedrobaeza: No, v11 is working fine as there's no collision in the column names.

eLBati: @pedrobaeza I was referring to `ail.price_unit` --> `AVG(ail.price_unit) AS price_unit`. Why is this change needed for v12?

pedrobaeza: Well, next time please put the comment in the proper line for correct referral. This is an improvement I have seen for putting the average price, because a sum on this field doesn't make sense IMO.

eLBati: There is not SUM in v11, just `ail.price_unit`. In v11 no aggregate function is used for `quantity`, `price_unit`, `price_subtotal` and `percentage`. This is why I don't understand how  this can work in v11.

Anyway, this PR is OK for me

eLBati (approved):

sale_commission/report/sale_commission_analysis_report.py
pedrobaeza: Ah, I remember then: in v11, the SQL view is grouping also by these fields, which is incorrect. It should be fixed in v11 then.

pedrobaeza:
@aitorbouzas any objection to merge this?

pedrobaeza:
Let's merge this!

PR #196, opened 2019-01-17
[FIX]sale_commission: Add security rules to avoid multicompany errors

javierjcf:
If you do a settlement, in a company A, and there are account.invoice.line.agent, in both company A and B (B is not a child company of A), When we do the settlements (by company) we need to avoid searching lines in the B company, because of security error reading field company_id of the related object. The next code in the wizard seattle, fail with multi company because of  the line 'lambda r: r.object_id.company_id == company.'

```
for company in agent_lines.mapped('company_id'):
    agent_lines_company = agent_lines.filtered(
        'lambda r: r.object_id.company_id == company)
```
Agent lines must be of the same or child company, never a brother company.

### DIFF
--- a/sale_commission/__manifest__.py
+++ b/sale_commission/__manifest__.py
@@ -18,6 +18,7 @@
     ],
     'data': [
         'security/ir.model.access.csv',
+        'security/commission_security.xml',
         'views/sale_commission_view.xml',
         'views/sale_commission_mixin_views.xml',
         'views/product_template_view.xml',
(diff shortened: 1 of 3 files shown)


etobella:
Why do you need to make it multicompany aware? This PR changes the behaviour of the commission a lot. Also, if it is required, it can give some problems on multicompany existing environments.

pedrobaeza:
I have made a soft modification that avoids the error but don't modify behavior through d14f1c2fed4d7c4c218498fb8da3b88b06832cd3, so I close this one.

PR #197, opened 2019-01-19
sale_commission: propagate agents from company to contacts

ernestotejeda:
Cc @Tecnativa

### DIFF
--- a/sale_commission/models/res_partner.py
+++ b/sale_commission/models/res_partner.py
@@ -1,3 +1,6 @@
+# Copyright 2016-2019 Tecnativa - Pedro M. Baeza
+# Copyright 2018 Tecnativa - Ernesto Tejeda
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
 
 from odoo import api, fields, models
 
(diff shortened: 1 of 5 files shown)


pedrobaeza (commented):
We can assign also `agents` on create when `'parent_id' in vals and agents` not in vals`

sale_commission/models/res_partner.py
```diff
@@ -36,3 +36,38 @@ class ResPartner(models.Model):
     def onchange_agent_type(self):
         if self.agent_type == 'agent' and self.agent:
             self.supplier = True
+
+    @api.onchange('parent_id')
+    def onchange_parent_id(self):
+        """Change Salesperson or Sales Channel if the parent company changes
+        and there's no Salesperson or Sales Channel defined yet"""
```
cubells: This docstring I think is incorrect.

cubells (changes requested):
You should increase the module version too.

ernestotejeda:
No problem when you create a contact from * Sales -> Orders -> Customers * and you select the parent company to which it belongs, but, when the contacts are created as part of the insertion of data for the creation of the parent company; From the contact's create method, the agents of the parent company (self.browse (vals ['parent_id']). agents) returns an empty record set, so I can not pass the company's agents to those contacts.
I do not know why self.browse(vals['parent_id']).agents returns an empty record set, apparently this happens because it's a many2many field.
I can't pass this field as default by context apparently because it's a many2many field

pedrobaeza:
@ernestotejeda check last commit. With this, it's working without problems.

pedrobaeza (approved):

ernestotejeda:
I added a known issue and squash

PR #198, opened 2019-01-22
[commission] Fix TypeError: not all arguments are converted during string formatting

robyf70:
I'm getting TypeError: not all arguments are converted during string formatting when assigning more than 1 agent to a sale order. This patch fixes the issue.


pedrobaeza:
This has nothing to do with the code, but probably with the translation. Please check that. Closing meanwhile.

PR #200, opened 2019-02-04
[11.0][FIX] avoid KeyError

dcorio:
These lines were breaking other modules inheriting this method with sale_commission installed

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -80,8 +80,10 @@ class AccountInvoiceLine(models.Model):
         """Add agents for records created from automations instead of UI."""
         # We use this form as this is the way it's returned when no real vals
         agents_vals = vals.get('agents', [(6, 0, [])])
-        if agents_vals and agents_vals[0][0] == 6 and not agents_vals[0][2]:
-            inv = self.env['account.invoice'].browse(vals['invoice_id'])
+        invoice_id = vals.get('invoice_id', False)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is a nonsense... `invoice_id` is a required field, so no possible to not indicate it.

dcorio:
@pedrobaeza damn...ignore it! it's AccountInvoiceLine and not AccountInvoice...sorry for the noise

dcorio:
@pedrobaeza after having ran a few checks I can say that:

1. invoice_id on account.invoice.line isn't mandatory
2. there are modules that first create the account.invoice.line and the account.invoice after
3. this is the only module so far that breaks the chain

If you still believe this is "nonsense" I'll try to create PR against the other repos with modules that creates account.invoice.lines records without passing invoice_id even if it's not mandatory.

pedrobaeza:
Hi @dcorio, I have checked in odoo itself and you're right:

https://github.com/odoo/odoo/blob/64f06858362acf584e4371e3848a6305ecea6b25/addons/account/models/account_invoice.py#L1485

They have removed the required flag from the line, so your PR is correct and apologies from my initial mistake.

pedrobaeza (approved):

warp10 (approved):
LGTM

pedrobaeza:
Cherry-picked for v12 in 410b28cbcbe1c845eb14d01e63d6b897f683a608

PR #201, opened 2019-02-12
[FIX] hr_commission : super call arguments

darshan-odoo:
Fixes super call argument for HR Employee


NL66278:
@darshan-odoo Hi, what is the reason for this change/fix? Did any problem occur? The more simple way to call super (not needing to specify the inheriting class) is an integral part of python 3.x.

darshan-odoo:
@NL66278 Yes right Ronald, I tested it now, it does work, was not aware of this explicit way, thanks!

Closing it!

PR #203, opened 2019-02-20
[12.0][MIG] sale_commission_pricelist

novawish:


sale_commission_pricelist/models/__init__.py
```diff
@@ -1,3 +1,3 @@
 
-from . import product_pricelist
+from . import product_pricelist_item
```
pedrobaeza: Well, the name was put as the main model in product.pricelist, not product.pricelist.item, so it contains possible changes to main model and also changes on children models.

pedrobaeza (approved):

sale_commission_pricelist/models/__init__.py
novawish: I'm not sure what you mean, does it mean that i should keep the old name?

pedrobaeza: Yes, not mandatory, but it was done that way by purpose, so no need of change it.

pedrobaeza:
You can also squash together both "[UPD] README.rst" adjacent commits

ernestotejeda:
Sorry, I made a PR for the same purpose (https://github.com/OCA/commission/pull/205) without realizing that this already existed.
Anyway, in the PR that I have done, I generated the readme, I fixed an error that was carried from the v11 (When the button 'Recompute lines agents' is clicked, it explodes) and I have added unit tests to cover the part of the code that caused the error.
Do you think we could close this PR to keep the other one with those additional changes?

yajo:
Oops sorry, I wasn't aware of this PR and I merged #205. Sorry for that!

Feel free to open a new PR if you need any improvements. Regards.

novawish:
@ernestotejeda  @Yajo Sure, no problem

PR #204, opened 2019-02-26
[IMP] sale_commission: move check of the amount to the end

etobella:
The check has been moved to the end in order to perform the join of a negative settlement and a positive settlement if possible.

### DIFF
--- a/sale_commission/__manifest__.py
+++ b/sale_commission/__manifest__.py
@@ -1,7 +1,7 @@
 # Copyright 2014-2018 Tecnativa - Pedro M. Baeza
 {
     'name': 'Sales commissions',
-    'version': '11.0.2.1.0',
+    'version': '11.0.2.1.1',
     'author': 'Tecnativa,'
               'Odoo Community Association (OCA)',
     'category': 'Sales Management',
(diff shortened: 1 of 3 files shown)


pedrobaeza (changes requested):
Please make an iterative check for breaking when the first exception is found instead of needing to compute whole set before raising the exception. A test with the user case (positive and negative) would be very advisable also

etobella:
You should wait to the end. Let's assume that an agent has two settlements, the first negative and the second positive and the total is positive. If we fail on the first iteration, it will fail, but with the current code it will wait until the end and check it later. If we want to check it inside the loop, we should sort the data by total.
Also, it cannot be checked here, as the module will create an invoice for each settlement. A second module is necessary in order to show it.

pedrobaeza:
I think you haven't understood me. I'm talking about:

1. Make a `for` loop instead of `mapped` + `filtered`.
2. Add unit tests.

etobella:
Ok! Sorry for the misunderstanding :wink:

etobella:
@pedrobaeza your comments have been attended.

pedrobaeza (approved):

JordiBForgeFlow (approved):
:+1: Tested on runbot. Tech review

PR #207, opened 2019-03-11
[10.0][IMP] sale_commission_formula: Compute commissions using unified method

SimoRubi:
The unified method being `_get_commission_amount`.
Before this PR, for each line all the lines in `self` were recomputed too


pedrobaeza:
Why this is not done in `sale.commission.line.mixin` object and remove the other 2 overrides (in invoice and in sales order)?

SimoRubi:
> Why this is not done in `sale.commission.line.mixin` object and remove the other 2 overrides (in invoice and in sales order)?

Thanks @pedrobaeza for the review, I didn't notice that there was the same code in `sale.order` too, done

sale_commission_formula/models/sale_commission_mixin.py
```diff
@@ -0,0 +1,29 @@
...
+    def _get_commission_amount(
+            self, commission, subtotal, commission_free, product, quantity):
+        self.ensure_one()
+        if commission.commission_type == 'formula' and \
+                not commission_free and commission:
+            formula = commission.formula
+            results = self._get_formula_input_dict()
+            safe_eval(formula, results, mode="exec", nocopy=True)
+            return float(results['result'])
+        else:
```
pedrobaeza: No need of `else`

sale_commission_formula/models/sale_commission_mixin.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import api, models
+from odoo.tools.safe_eval import safe_eval
+
+
+class SaleCommissionMixin(models.AbstractModel):
+    _inherit = 'sale.commission.mixin'
+
+    @api.model
+    def _get_formula_input_dict(self):
+        return {'line': self.object_id,
```
pedrobaeza: Why do you overwrite this?

pedrobaeza (changes requested):

sale_commission_formula/models/sale_commission_mixin.py
SimoRubi: @pedrobaeza I was right now testing this locally, I have overridden `sale.commission.mixin` by mistake, the correct model to override is `sale.commission.line.mixin`, fixed in https://github.com/OCA/commission/compare/6e24d9bad74ee98e56cba5be6ee4a95f43865341..6e998cdfb1e42ba552e154bb4e92b6033376a808

sale_commission_formula/models/sale_commission_mixin.py
SimoRubi: @pedrobaeza I am calling super to get the correct `amount` if the `commission_type` is not `formula`.

sale_commission_formula/models/sale_commission_line_mixin.py
```diff
@@ -0,0 +1,29 @@
...
+    def _get_commission_amount(
+            self, commission, subtotal, commission_free, product, quantity):
+        self.ensure_one()
+        if commission.commission_type == 'formula' and \
+                not commission_free and commission:
+            formula = commission.formula
+            results = self._get_formula_input_dict()
+            safe_eval(formula, results, mode="exec", nocopy=True)
+            return float(results['result'])
+        else:
```
pedrobaeza: What I mean is that you can put `super` call without the else.

sale_commission_formula/models/sale_commission_line_mixin.py
```diff
@@ -0,0 +1,29 @@
...
+
+from odoo import api, models
+from odoo.tools.safe_eval import safe_eval
+
+
+class SaleCommissionLineMixin(models.AbstractModel):
+    _inherit = 'sale.commission.line.mixin'
+
+    @api.model
+    def _get_formula_input_dict(self):
```
pedrobaeza, line 13: Again, this shouldn't be needed

sale_commission_formula/models/sale_commission_line_mixin.py
SimoRubi: Oh, ok I didn't get that, done

sale_commission_formula/models/sale_commission_line_mixin.py
SimoRubi: I'm sorry but I'm not getting what you mean here.
The method is created to be overridden from whoever might be willing to add values for the formula to use. 
I can't find any other occurrence of `_get_formula_input_dict` so it is not overriding anything

SimoRubi: On a side note, I just edited that to use the correct decorator `api.multi` with `ensure_one`

SimoRubi:
Travis's automated tests have an error not depending on this, to be fixed in https://github.com/OCA/commission/pull/208

sale_commission_formula/models/sale_commission_line_mixin.py
pedrobaeza: OK, got it. I thought this method was declared already on super. My fault.

pedrobaeza (approved):

SimoRubi:
Rebased after #208

eLBati (approved):

PR #208, opened 2019-03-12
[10.0][FIX] sale_comission: Fix tests

SimoRubi:
sale_order4 is using agent with 'annual' commission, so in order to have settlements, we have to call the wizard with date_to equal to next year


pedrobaeza:
Fast-tracking this.

P.S.: That's why I always prefer to not use demo data and populate everything in tests, for having all data under the same scope.

SimoRubi:
@pedrobaeza that is one good reason, anyway when lots of data should be created, in my opinion the use of `demo_data` is preferrable, in order to have a more readable code

pedrobaeza:
You don't gain too much on readability in XML data. As everything is populated in setUp, readability doesn't suffer, and if you want, you can have a base test file with this population for splitting things, but I prefer to not do it.

PR #216, opened 2019-05-20
[10.0][IMP] sale_commission: Add settlement lines' analysis views

SimoRubi:
Add menu item for analysis views:

- graph: declared but unused
- pivot: new view, similar to graph


pedrobaeza:
Have you seen the work about analysis done in 11.0/12.0? I think it's more interesting to backport it than to create a divergence in views/features that will difficult migration path.

SimoRubi:
No i'm mostly using 10.0. Ok I'll take a look and backport a version for an easier migration path. 11.0 or 12.0?

pedrobaeza:
Both are the same basically. You can cherry-pick d11095d76b25d8308931bacff6ff971af2593335 and 64dc262a10a408c10e54a1cf3e36b74289dc7bc4

SimoRubi:
Thanks @pedrobaeza for the hint, I tried to have a more complete set of the new commits with
`git log origin/10.0..origin/11.0 --no-merges -- sale_commission`
but it was producing all the commits n 11.0i because migrating creates new commits, if you have a better way to do that I'll be happy to try the backporting.

As for this PR, I cherry-picked as you suggested.

sale_commission/models/settlement.py
```diff
@@ -33,6 +34,11 @@ def _default_currency(self):
     currency_id = fields.Many2one(
         comodel_name='res.currency', readonly=True,
         default=_default_currency)
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        default=lambda self: self.env.user.company_id,
+        required=True
+    )
```
eLBati, line 41: Shouldn't `company_id` be used in a record rule?

SimoRubi: @eLBati it could be added in order to add the 'automatic filtering' in multi-company environments but, as far as I know, it is not mandatory.
In particular, I added this field because it was necessary after the changes backported from v11 and there (https://github.com/OCA/commission/tree/11.0/sale_commission) I haven't found any record rule.

pedrobaeza: I think we should add that record rules for proper multi-company support, which is indeed a pending topic.

SimoRubi: ok then I'll add them in a separate commit so we can easily forward port them

pedrobaeza: OK, great. You should put a migration scripts for emptying `company_id` field for avoiding mis-behaviors when you have commissions in several companies, or even better, try an heuristic for deducing the company of each settlement.

SimoRubi: @pedrobaeza emptying `company_id` is not possible because it is `required`, I'm trying to find the correct company for each settlement

SimoRubi: @pedrobaeza @eLBati I added the record rules and the migration scripts as you requested, please have a look, thanks

sale_commission/migrations/10.0.2.4.0/pre-populate_company.py
```diff
@@ -0,0 +1,23 @@
...
+        'WHERE sale_commission_settlement.agent = res_partner.id')
+
+    # Populate company_id for sale.commission.settlement.line,
+    # get the company from the settlement
+    cr.execute(
+        'UPDATE sale_commission_settlement_line '
+        'SET company_id = sale_commission_settlement.company_id'
+        'FROM sale_commission_settlement'
+        'WHERE sale_commission_settlement_line.settlement '
+        '   = sale_commission_settlement.id')
```
eLBati: Can this be named `pre-populate_company.py`? Did you test it?

pedrobaeza: Yes, actually it can be named `pre-<whatever>` for being executed.

tafaRU: AFAIK the only requirement for ensuring the script migrations will be executed is that the python file names must start by `pre` or `post`.

pedrobaeza: Actually it's `pre-` / `post-` (with dash).

SimoRubi: > Actually it's `pre-` / `post-` (with dash).

Correct, from https://github.com/odoo/odoo/blob/7746621b92c27f6250c0895b0c54b501097e05bb/odoo/modules/migration.py#L28

sale_commission/models/settlement.py
pedrobaeza: I'm seeing that the company is already present in 11.0: https://github.com/OCA/commission/blob/0661addc5f0d8e28840e68d637f6297b8a85ac41/sale_commission/models/settlement.py#L37

sale_commission/models/account_invoice.py
```diff
@@ -117,6 +117,15 @@ class AccountInvoiceLineAgent(models.Model):
     settled = fields.Boolean(
         compute="_compute_settled",
         store=True, copy=False)
+    company_id = fields.Many2one(
```
pedrobaeza, line 120: Is it needed to add this burden?

sale_commission/migrations/10.0.2.4.0/pre-populate_company.py
pedrobaeza: I know it because I was the one that needed to correct that in Odoo documentation: https://github.com/odoo/odoo/commit/c0e35f84e23ebda8ea6033711ded5638ef309da2 :wink:

SimoRubi: > Can this be named `pre-populate_company.py`? Did you test it?

I tested the queries in postgresql, did not test the migration execution.
Now I have tried this and it says
```
ProgrammingError: column "company_id" of relation "sale_commission_settlement_line" does not exist
LINE 1: UPDATE sale_commission_settlement_line SET company_id = sale...
```

Maybe the script should be a post-<stuff>?

pedrobaeza: Yes, as the column is populated on the update itself (after pre and before post).

SimoRubi: Changed to post- and error was still there, then removed the query on the lines because I hadn't noticed that it is a related field (non-stored by default)

sale_commission/models/account_invoice.py
SimoRubi: now checking

SimoRubi: It is needed otherwise:
```
commission/sale_commission/models/settlement.py", line 198, in _check_company
    if record.agent_line.company_id != record.company_id:
AttributeError: 'account.invoice.line.agent' object has no attribute 'company_id'
```
I added it in its commit message.
Being here `company_id` a computed field I don't think any migration script is needed, but I've added a record rule

pedrobaeza: The problem I see is performance, so balance the need (or other way of getting the same) with this in mind.

sale_commission/migrations/10.0.2.4.0/pre-populate_company.py
eLBati: @SimoRubi So is it working as expected according to your tests now?

sale_commission/models/account_invoice.py
SimoRubi: @pedrobaeza would a related field be more efficient in your opinion?
Because looking at the computing method, that should be enough.

On a side note, the computing method is exactly the same as it is in the following versions, there might be performance issues over there too

pedrobaeza: Well, I was just guessing... if there's a record rule directly associated to this model, then the related stored field is the best option between performance and feature.

sale_commission/migrations/10.0.2.4.0/pre-populate_company.py
SimoRubi: Yes in my local environment it is working as expected it is @eLBati

sale_commission/models/account_invoice.py
SimoRubi: @pedrobaeza changed to related field

sale_commission/models/account_invoice.py
```diff
@@ -117,6 +117,14 @@ class AccountInvoiceLineAgent(models.Model):
...
+        related="object_id.company_id",
+        store=True,
```
pedrobaeza, line 122: `readonly=True`

SimoRubi: thanks, added

eLBati (approved):
LGTM

@SimoRubi do you want to keep the last 3 commits separated?

SimoRubi:
> LGTM
> 
> @SimoRubi do you want to keep the last 3 commits separated?

Yes, in my opinion they refer to different improvements that might be forward ported more easily if they are separated commits

Issue #215, opened 2019-04-29
Agent assignation invisible in contact > Sales n Purchase


damsoi01:
Hi,
I installed the module for odoo 12, in contacts > Sales & Purchase I can only check the box to make this contact an agent or not.
In odoo 10 we used to have a field that could allow us to assign by default an agent so when creating an order we do not need to do it manually on each product.
How can I get the buyers to be assigned to a specific agent by default?
<img width="1154" alt="Screen Shot 2019-04-29 at 12 56 59 PM" src="https://user-images.githubusercontent.com/17989353/56919623-4d7d6d80-6a7e-11e9-8f03-7204d1f5810b.png"> No field to assign agent in this example

Thank you!

tpascualwottan:
Hello, I'm not a expert, but I miss an xpath for Agents field in /views/res_partner_view

pedrobaeza:
Fixed in #217

PR #217, opened 2019-06-17
[12.0][FIX] sale_commission: show agents field on partner form

rubencr7:
This commit fixes the issue #215


pedrobaeza (approved):
Fast-tracking this for fixing the problem

Issue #35, opened 2015-10-26
Commissions for Website Sales


Cywaithaka:
module: sale_commission
version: 8.0

**Current behavior**
When a sales order/quotation is created on the backend, the commissions are duly calculated. However, when the same customer places  an order to on the ecommerce shop, commissions are zero.
**Expected behavior**
1.If Customer1 places an order on the website, the commissions should be calculated just like when the  order is placed from the backend.

pedrobaeza:
For now that option is not taking into account, but it would be easy to intercept method that creates the sale order line and inject commission values. Maybe you can contribute with this _website_sale_commission_ module.

Cywaithaka:
I tried extending the _website_product_id_change function in sale_order with no success. 

``` python
class SaleOrder(osv.Model):
    _inherit = "sale.order"
    def _website_product_id_change(self, cr, uid, ids, order_id, product_id, qty=0, line_id=None, context=None):
        res = super(SaleOrder,self)._website_product_id_change(cr, uid, ids, order_id, product_id, qty=qty, line_id=line_id, context=None)
        line = self.pool.get('sale.order.line').browse(cr, SUPERUSER_ID, line_id, context=context)
        res['commissions'] = line._default_commissions()
        return res
```

Since the function that populates the commissions is called as a default value, could there be a reason why this default does not apply when the sale order line is created from the website? That is, using the sale_commission module as it is.

pedrobaeza:
I think it should be `line._default_agents`, but I haven't tried.

Cywaithaka:
I have created a repo for the module here https://github.com/Cywaithaka/website_sale_commission/tree/master. Hope we can have other people to contribute. What I have tried so far has not worked.

Cywaithaka:
Hi @pedrobaeza Apologies, I was using @savoirfairelinux branch (8.0_sale_commission_multiple) from here https://github.com/savoirfairelinux/commission/tree/8.0_sale_commission_multiple/sale_commission and I notice there are some naming differences with the main branch. This is why I used line._default_commissions

pedrobaeza:
Closing the same as the PR as explained there.

PR #218, opened 2019-06-19
[10.0][ADD] website_sale_commission

ozydev:
Wrote an addon to override the _website_product_id_change of website_sale addon and add agents on sale order line.
Connection with https://github.com/OCA/commission/issues/35


pedrobaeza:
Is this still needed with latest modifications introduced in the module? As it's anyway red, I'm closing till any answer.

PR #219, opened 2019-07-01
[12.0][MIG] hr_commission

kongrattapong:
done.


hr_commission/models/res_partner.py
```diff
@@ -11,6 +11,7 @@ class ResPartner(models.Model):
         selection_add=[("salesman", "Salesman (employee)")],
     )
     employee_id = fields.Many2one(
+        string="Employee.",
```
pedrobaeza, line 14: Remove the final dot

hr_commission/__manifest__.py
```diff
@@ -3,10 +3,10 @@
 
 {
     'name': 'HR commissions',
-    'version': '11.0.1.0.0',
+    'version': '12.0.1.0.0',
     'author': 'Tecnativa,'
               'Odoo Community Association (OCA)',
-    "category": "Commissions",
+    'category': 'Commissions',
```
pedrobaeza: Better to avoid change quote style

pedrobaeza (changes requested):

hr_commission/models/res_partner.py
kongrattapong: We got this problem so we need to add dot to make a different
```
2019-07-01 06:18:34,600 13216 WARNING odoo12_hr_commission odoo.addons.base.models.ir_model: Two fields (employee_id, employee) of res.partner() have the same label: Employee. 
2019-07-01 06:18:34,636 13216 WARNING odoo12_hr_commission odoo.addons.base.models.ir_model: Two fields (employee_id, employee) of res.users() have the same label: Employee. 
```

hr_commission/__manifest__.py
kongrattapong: Okay.

hr_commission/models/res_partner.py
pedrobaeza: And where does `employee` come from?

kongrattapong: employee from res_partner.py in module base.

https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_partner.py#L180

pedrobaeza: OK, that's a new field. We can call this one then `Employee reference`, but I would beg you to check a bit deeper now if there's another link in standard that makes the same function that this `employee_id` field.

kongrattapong: I checked and found nothing.

pedrobaeza: OK, thanks, then the other label is better IMO

kongrattapong: @pedrobaeza thank for your reviews.

kongrattapong:
Does this PR have to change anything more?

pedrobaeza:
You have to comment when changes are done for noticing to reviewers that they can re-review, or it will be unnoticed.

Please squash migration commits into one (https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests), and tell me back.

kongrattapong:
> You have to comment when changes are done for noticing to reviewers that they can re-review, or it will be unnoticed.
> 
> Please squash migration commits into one (https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests), and tell me back.

@pedrobaeza  Oh sorry, I squash already

pedrobaeza (approved):

pedrobaeza:
As this is very straight-forward, let's merge it:

/ocabot merge

PR #221, opened 2019-08-19
[10.0][FIX] sale_commission: Add view context by code

pedrobaeza:
As Odoo hasn't got any mechanism for adding several elements to the context dict, using XML view tools, we can have our context replaced by other module doing the same operation.

Example: https://github.com/OCA/sale-workflow/blob/8428446c79981aee281eb1773ff84c5a32d02067/sale_order_line_sequence/views/sale_view.xml#L13

Thus, we avoid the problem adding the context by code, as this context is vital for having default agents commissions populated on the sales order lines.

cc @Tecnativa

### DIFF
--- a/sale_commission/models/sale_order.py
+++ b/sale_commission/models/sale_order.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
-# Copyright 2014-2018 Tecnativa - Pedro M. Baeza
+# Copyright 2014-2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
+from lxml import etree
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@SimoRubi can you check this?

rafaelbn (approved):
Functionally :+1:

SimoRubi:
Hi, this is indeed a bad issue due to the current implementation of Odoo.
A permanent fix should be implemented in Odoo core, do you know if there is any opened issue about it? At least we could give it a :+1: to let them know it's still on

I have given a few tries to the `add` attribute of the `attribute` node but it is not very simple and the implementation would be overkilling the problem.

The implementation you provided is indeed robust, but I find that the idea behind it may have some flaws, for instance: what if other modules add a context using an xpath different from `//field[@name='order_line']`?

As long as I know, this problem is usually solved by a *third* auto-installing module that manages the conflict, so I suggest this solution.

pedrobaeza:
> A permanent fix should be implemented in Odoo core, do you know if there is any opened issue about it? At least we could give it a +1 to let them know it's still on

I'm looking for having it since a long... see https://github.com/odoo/odoo/pull/26607

> The implementation you provided is indeed robust, but I find that the idea behind it may have some flaws, for instance: what if other modules add a context using an xpath different from //field[@name='order_line']?

This is not a problem, as this is only for getting to the node. Once located, the context is added transparently, no matter what others modules have done with it.

> As long as I know, this problem is usually solved by a third auto-installing module that manages the conflict, so I suggest this solution.

No, this is the worst solution, uselessly overheading with lots of modules and possible combinations (imagine if you have 3 modules A, B and C modifying same context - you would need glue_A_B, glue_A_C, glue_B_C, glue_A_B_C). The ideal solution is that all modules modifying context use this same technique.

SimoRubi (approved):
This is a work-around so not the best way to do it (that would be https://github.com/odoo/odoo/pull/26607).
Anyway I can't see any better way to work-around this issue.

SimoRubi:
I'm not sure if I exactly know the idea behind https://github.com/OCA/OCB, but isn't this the kind of thing that could be added in there?

pedrobaeza:
Definitively not. That would create modules that are only compatible with OCB and not with vanilla Odoo.

pedrobaeza:
Cherry-picked for v11 in 75265c7daf566fb96a466955fe7f37181327c349 and for v12 in 1578ff03bb8ff62db42ff07ad39d8d28add69484

PR #223, opened 2019-09-13
[FIX]sale_commission: Bug al imprimir informe comisiones

javierjcf:
When you print the settlement report you get a bug.
In v10 internal_layout and html_conmtainer are in report module.


pedrobaeza (approved):
Bad backport of the feature

daramousk (approved):

PR #231, opened 2020-01-15
[12.0][IMP] sale_commission: move check of the amount to the end (#204)

jarroyomorales:
Port of https://github.com/OCA/commission/pull/204 to v12.

cc @etobella


pedrobaeza (approved):
Fast-tracking as it's a cherry-pick

PR #232, opened 2020-02-13
[12.0][IMP] sale_commission: Show signed subtotal

SimoRubi:
In the commission analysis report, invoice lines coming from a refund have a negative commission amount (correct) and a positive subtotal (can be misleading).

After this PR, both amounts will be negative for refunds and positive for invoices.


tafaRU (approved):
LGTM, thanks!

pedrobaeza (changes requested):
Please don't remove any field in stable, as you are potentially breaking other modules that depend on that data. Just add the new field. You can change as well the default shown measure in the action window, but not to remove the previous field.

SimoRubi:
Thanks @pedrobaeza for the review, ok noted: changing existing fields -> bad

pedrobaeza (approved):

PR #243, opened 2020-06-12
[11.0][ADD] sale_commission_salesman: Sales commissions from salesman

pedrobaeza:
It includes an extra commit that refactors `sale_commission` module for making possible the other module.

Sales commissions from salesman
==========================

This module extends the functionality of `sale_commission` to allow you to auto-populate salesmen as commission agents without setting explicitly them on the customer.

**IMPORTANT**: The salesman will be only populated if no other commission agent is set via other method, to assure that there's no overlapping commissions.

Configuration
=============

* Go to *Contacts*.
* Search for the contact linked to the salesman user and open it.
* If it's actually linked to a user, you'll see a new check in the *Agent information* page, called "Convert salesman into agent".
* Check it for having that salesman automatically populated as agent. You need to assign a commission to that agent as well.

Usage
=====

To use this module, you need to:

* Go to *Sales > Orders > Quotations* and create a new one.
* Fill the customer with one with no agents set.
* Select a salesman with the mark checked.
* Add a line in the quotation, and you will see the salesman populated as commission agent.

cc @Tecnativa TT20994


carlosdauden (approved):

chienandalu (approved):
Tested in runbot :+1:

sale_commission/models/sale_commission_mixin.py
```diff
@@ -66,11 +66,12 @@ def _compute_commission_status(self):
                     "%s commission agents"
                 ) % len(line.agents)
 
-    def _prepare_agents_vals(self):
+    def _prepare_agents_vals(self, vals=None):
         """Hook method for preparing the values of agents.
 
         :param: self: Record of the object that is being handled.
         """
+        assert self or vals
```
rousseldenis, line 74: Use orm exception instead

pedrobaeza: This is a developer assert, not a user one. That's why I put it that way.

PR #248, opened 2020-07-11
[13.0] [MIG] sale_commission_formula

feketemihai:


sale_commission_formula/models/sale_commission.py
```diff
@@ -13,6 +13,6 @@ class SaleCommission(models.Model):
         "Formula",
         default="if line._name == 'sale.order.line':\n"
         "    result = 0\n"
-        "if line._name == 'account.invoice.line':\n"
+        "if line._name == 'account.move.line':\n"
```
pedrobaeza, line 16: You need a migration scripts that rewrites all the existing formulas replacing this.

sale_commission_formula/README.rst
```diff
@@ -14,16 +14,16 @@ Sale Commission Formula
...
-    :target: https://translation.odoo-community.org/projects/commission-12-0/commission-12-0-sale_commission_formula
+    :target: https://translation.odoo-community.org/projects/commission-13-0/commission-13-0-sale_commission_formula
     :alt: Translate me on Weblate
 .. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
-    :target: https://runbot.odoo-community.org/runbot/165/12.0
+    :target: https://runbot.odoo-community.org/runbot/165/13.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| |badge5| 
+|badge1| |badge2| |badge3| |badge4| |badge5|
```
pedrobaeza: It seems you have edited this by hand, isn't it?

pedrobaeza (changes requested):

sale_commission_formula/models/sale_commission.py
feketemihai: @pedrobaeza Done.

sale_commission_formula/README.rst
feketemihai: Done

pedrobaeza (approved):

tafaRU (changes requested):
.pre-commit-config.yaml changes shouldn't be done in a separate PR?

The rest LGTM. Thanks!

pedrobaeza:
@tafaRU not really, as they are required for this to be correct, so it's OK to include at the same time if we handle it quickly. Obviously, having in a separate PR can make it to be fast-tracked, and then avoid possible merge conflicts (which we will have with the other module done be Fekete).

feketemihai:
@pedrobaeza @tafaRU you can merge one and I will fix conflicts if it appears in the second PR.

tafaRU:
> @tafaRU not really, as they are required for this to be correct, so it's OK to include at the same time if we handle it quickly. 

Ok thanks @pedrobaeza I approve then.

tafaRU (approved):

PR #249, opened 2020-07-11
[13.0] [MIG] sale_commission_pricelist

feketemihai:


sale_commission_pricelist/models/sale_order.py
```diff
@@ -26,13 +26,12 @@ def _get_commission_from_pricelist(self):
     def _onchange_product_id_sale_commission_pricelist(self):
```
pedrobaeza: This should be replaced by a new `depends` on the compute field of `agent_ids` to `order_id.pricelist_id`.

pedrobaeza (changes requested):
Please squash both migration commits (the pylint one and the real migration).

sale_commission_pricelist/models/sale_order.py
feketemihai: @pedrobaeza Done

pedrobaeza (approved):

feketemihai:
@pedrobaeza Merge 13.0 again, so no conflict.

carlosdauden (approved):

PR #252, opened 2020-07-27
[12.0][MIG] sale_commission_salesman: Migration to 12.0

pedrobaeza:
- Standard procedure
- It includes the needed refactoring in `sale_commission` module done in 11.0, in a separate commit.

@Tecnativa TT24802

PR #256, opened 2020-08-03
[11.0][FIX] sale_commission: No commission should be populated on vendor bills

alessandrocamilli:
Backport of https://github.com/OCA/commission/commit/41a90c4d8654c1284484602b55bf56fa436d9884


pedrobaeza:
It seems this breaks a test in `sale_commission_salesman`. Please check differences between both branches 11.0 and 12.0 (with #252).

PR #266, opened 2020-11-13
[13.0][FIX] sale_commission: fixing error on append instruction

dennybiasiolli:
without this fix, during the migration appears this error:
`TypeError: append() takes exactly one argument (4 given)`

### DIFF
--- a/sale_commission/migrations/13.0.1.0.0/pre-migration.py
+++ b/sale_commission/migrations/13.0.1.0.0/pre-migration.py
@@ -39,5 +39,5 @@ def migrate(env, version):
     renamed_fields = []
     for model in renamings:
         for old_field, new_field in renamings[model]:
-            renamed_fields.append(model, model.replace(".", "_"), old_field, new_field)
+            renamed_fields.append((model, model.replace(".", "_"), old_field, new_field))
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this:

/ocabot merge nobump

pedrobaeza:
@dennybiasiolli I have pushed directly the fix, as black reformats the code.

dennybiasiolli:
@pedrobaeza I think there is an error, because `_compute_commission_amount` does not exists
https://github.com/OCA/commission/commit/1ea3a499c757d848fbdcf7fd110f07a1cfa3e8d2#diff-58468815aafb2137f4d5de9c163c3b69fee52d2d231d147c606b5f8478966e03R119

pedrobaeza:
Yeah, I have just seen it. I had in my branch that change and commit all together. Fixed in https://github.com/OCA/commission/commit/05fb8b49278e248be8ac070c870dd75c49e9f653

Sorry for the inconveniences and thanks for noting.

dennybiasiolli:
No problem, thanks for the fix! 😉

PR #267, opened 2020-11-16
[13.0][FIX] sale_commission: fixing view_res_partner_filter

dennybiasiolli:
`<filter name="supplier">` was removed in https://github.com/odoo/odoo/commit/3e97cff7797c4f7e695d20def6b920561b6a6525

`<filter name="inactive">` is the first available field

### DIFF
--- a/sale_commission/views/res_partner_view.xml
+++ b/sale_commission/views/res_partner_view.xml
@@ -64,10 +64,10 @@
             </page>
         </field>
     </record>
-    <record id="view_res_partner_filter" model="ir.ui.view">
+    <record id="res_partner_view_search" model="ir.ui.view">
         <field name="name">res.partner.select</field>
         <field name="model">res.partner</field>
-        <field name="inherit_id" ref="base.view_res_partner_filter" />
+        <field name="inherit_id" ref="account.res_partner_view_search" />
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
Thanks for checking this. Indeed, this is not correct, but not due to the field, but the inheritance. Please inherit from `account.res_partner_view_search`, that is the one containing such field.

dennybiasiolli:
@pedrobaeza yeah right, your suggestion it's the better solution, thanks 😄

sale_commission/views/res_partner_view.xml
```diff
@@ -67,7 +67,7 @@
     <record id="view_res_partner_filter" model="ir.ui.view">
```
pedrobaeza: For following guidelines

```suggestion
    <record id="res_partner_view_search" model="ir.ui.view">
```

pedrobaeza (approved):
Thanks

pedrobaeza:
Ouch, it seems `sale_commission.view_res_partner_filter` (old search view name) was referred in other part of the code. Please locate it and replace it.

pedrobaeza:
Now it is!

Fast-tracking this then:

/ocabot merge minor

PR #270, opened 2020-12-03
[IMP] sale_commission: Remove required attribute in settlement field.Added to field in view form

sergio-teruel:
cc @Tecnativa TT27096
ping @Yajo @chienandalu

### DIFF
--- a/sale_commission/models/res_partner.py
+++ b/sale_commission/models/res_partner.py
@@ -18,7 +18,7 @@ class ResPartner(models.Model):
         string="Creditor/Agent",
         help="Check this field if the partner is a creditor or an agent.")
     agent_type = fields.Selection(
-        selection=[("agent", "External agent")], string="Type", required=True,
+        selection=[("agent", "External agent")], string="Type",
         default="agent")
     commission = fields.Many2one(
         string="Commission", comodel_name="sale.commission",
@@ -30,7 +30,7 @@ class ResPartner(models.Model):
                    ("quaterly", "Quarterly"),
                    ("semi", "Semi-annual"),
                    ("annual", "Annual")],
-        string="Settlement period", default="monthly", required=True)
+        string="Settlement period", default="monthly")
     settlements = fields.One2many(
         comodel_name="sale.commission.settlement", inverse_name="agent",
         readonly=True)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please remove the required to `agent_type` field as well.

sergio-teruel:
@pedrobaeza Changes done!!

bodedra:
@sergio-teruel can you explain the issue?

pedrobaeza (approved):

sergio-teruel:
Sorry @bodedra I think that @Yajo can explains better than me...

victoralmau (approved):

chienandalu (approved):

yajo (commented):
The code is perfect, but IMHO you need to rewrite the commit.

The commit says what you do, but doesn't say why. If you don't explain that, then in the future somebody will come (and that somebody might even be myself who discovered the bug but have some amnesia :laughing:) and revert this, because the reason is not obvious and, at first sight, the code looks "worse" than it was.

So, please just amend your commit with a meaningful message and this shall be ready to merge :blush:

yajo:
> Sorry @bodedra I think that @Yajo can explains better than me...

I just read this. :facepalm: Let me do the amend.

yajo:
Done. Reason: When updating a table, modules that don't have the `required=True` setting of a field in their scope issue an `ALTER TABLE` at least twice per module. This blocks the entire table.

This module was doing that, and since the `res.partner` model is so ubiquitous, this was blocking concurrent updates.

By moving the required attribute to the views, the UX is the same, but concurrent updates work fine again.

pedrobaeza:
Please fw-port it to v13

yajo:
Done in https://github.com/OCA/commission/pull/271

bodedra:
> Done. Reason: When updating a table, modules that don't have the `required=True` setting of a field in their scope issue an `ALTER TABLE` at least twice per module. This blocks the entire table.
> 
> This module was doing that, and since the `res.partner` model is so ubiquitous, this was blocking concurrent updates.
> 
> By moving the required attribute to the views, the UX is the same, but concurrent updates work fine again.

Thank you @Yajo for explanation. That's what I am looking for.

PR #271, opened 2020-12-04
[IMP] sale_commission: Remove required attribute in settlement field. Added to view

yajo:
When updating a table, modules that don't have the `required=True` setting of a field in their scope issue an `ALTER TABLE` at least twice per module. This blocks the entire table.

This module was doing that, and since the `res.partner` model is so ubiquitous, this was blocking concurrent updates.

By moving the required attribute to the views, the UX is the same, but concurrent updates work fine again.

@Tecnativa TT27096
FWport of https://github.com/OCA/commission/pull/270

### DIFF
--- a/sale_commission/models/res_partner.py
+++ b/sale_commission/models/res_partner.py
@@ -25,10 +25,7 @@ class ResPartner(models.Model):
         help="Check this field if the partner is a creditor or an agent.",
     )
     agent_type = fields.Selection(
-        selection=[("agent", "External agent")],
-        string="Type",
-        required=True,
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking as approved in v12 and it's the same:

/ocabot merge minor

pedrobaeza:
@Yajo please check pre-commit

yajo:
Done, sorry.

marcelsavegnago (approved):

PR #277, opened 2021-02-04
[IMP] sale_commission: Add hook for partner

etobella:
This PR adds a small hook in order to choose the partner.

@pedrobaeza

### DIFF
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -72,9 +72,13 @@ class Settlement(models.Model):
             'context': {'settlement_ids': self.ids}
         }
 
+
+    def _get_settlement_invoice_partner(self):
+        return self.agent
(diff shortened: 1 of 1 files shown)


sale_commission/models/settlement.py
```diff
@@ -72,9 +72,13 @@ def action_invoice(self):
             'context': {'settlement_ids': self.ids}
         }
 
+
```
pedrobaeza, line 75: Extra line

pedrobaeza (changes requested):
Anyway, you don't need any hook. Just override `_prepare_invoice_header` in your module and change `partner_id` in the returned dictionary (and any other possible field linked to it)

pedrobaeza:
Closing as not needed.

PR #278, opened 2021-02-04
[ADD] sale_commission_delegated_partner

etobella:
This module allows to change the partner of the invoice using a selected from the partner.
It might be interesting if we cannot use the parent_id as the agent might loose some information on that case

@pedrobaeza


sale_commission_delegate_partner/models/sale_commission_settlement.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2021 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+
+
+class SaleCommissionSettlement(models.Model):
+    _inherit = 'sale.commission.settlement'
+
+    def _get_settlement_invoice_partner(self):
```
pedrobaeza: Remove this

pedrobaeza (changes requested):
It should be sale_commission_delegated_partner following https://github.com/OCA/vertical-association/tree/12.0/membership_delegated_partner and override `_prepare_invoice_header` instead of adding another hook for the same

sale_commission_delegate_partner/models/sale_commission_settlement.py
etobella: If I make the change on `_prepare_invoice_header` I need to do again the new and the onchange and I thought it was better to avoid that. WDYT?

etobella:
Changed according to your comments :wink:

pedrobaeza (approved):

marcelsavegnago (approved):

PR #279, opened 2021-02-08
[12.0] sale_commission: add analysis view for sale orders

eLBati:


pedrobaeza:
What I miss here:

- Record rules for applying same permissions as in sale.order for "User: See own documents only".
- This is something not very valid, as final settlement depends on invoices, not on sales orders, so you can have a very old sales order, but not being invoiced until a lot, and the amounts will be distorted by this. My idea was to bring invoice data here to Sales section, but with the corresponding permissions.

eLBati:
> Record rules for applying same permissions as in sale.order for "User: See own documents only".

Thanks, done

> This is something not very valid, as final settlement depends on invoices, not on sales orders, so you can have a very old sales order, but not being invoiced until a lot, and the amounts will be distorted by this

The purpose of the analysis based on sales is  more about forecast indeed; basically: how much will the company have to pay, based on (confirmed) orders?

pedrobaeza:
Well, as said, that forecast can be very distorted due to the differences in dates from quotation/sales order to invoice.

eLBati:
Even if distorted, it is anyway useful information that our customers ask for (consciously of the possible distortion)

pedrobaeza:
OK, just add the information on the README with this warning (and thanks for the contribution). FYI, I have budgeted to a customer for the other information I'm talking about.

eLBati:
@pedrobaeza added, thanks

eLBati:
Another minor functionality added: 75ae859e4a6f41d939905a1cace6171477e3dccb

pedrobaeza:
Please split each feature in a PR. For the second one, please use a computed searchable field (with `search` method) instead of storing it, as it will be slow to update the module if not, and it's a waste of records for a non usual thing.

eLBati:
Moved the second commit to https://github.com/OCA/commission/pull/280

sale_commission/report/sale_commission_analysis_report.py
```diff
@@ -107,3 +107,104 @@ def init(self):
                 AsIs(self._group_by())
             ),
         )
+
+
+class SaleOrderCommissionAnalysisReport(models.Model):
```
pedrobaeza: Better to put this on a new file, as they are separate models.

sale_commission/report/sale_commission_analysis_report_view.xml
```diff
@@ -63,9 +63,74 @@
         <field name="view_mode">pivot,graph</field>
         <field name="search_view_id" ref="view_sale_commission_analysis_search"/>
         <field name="view_id" ref="view_sale_commission_analysis_graph"/>
-        <field name="help">This report performs analysis on your commissions added in invoice lines. You can check the amount and lines that will be settled by date, product, product category, aget, etc. Use this report to perform analysis on invoice lines agents not beeing settled yet.</field>
+        <field name="help">This report performs analysis on your commissions added in invoice lines. You can check the amount and lines that will be settled by date, product, product category, agent, etc. Use this report to perform analysis on invoice lines agents not beeing settled yet.</field>
     </record>
 
     <menuitem action="action_sale_commission_analysis_report" id="menu_sale_commissions_analysis" parent="account.account_reports_management_menu" sequence="15"/>
 
+    <record id="view_sale_order_commission_analysis_pivot" model="ir.ui.view">
```
pedrobaeza: The same for this.

sale_commission/security/ir.model.access.csv
```diff
@@ -14,3 +14,4 @@ access_sale_commission_settlement_line_manager,access_sale_commission_settlement
 access_sale_commission_settlement_line_salesman,access_sale_commission_settlement_line_salesman,model_sale_commission_settlement_line,sales_team.group_sale_salesman,1,0,0,0
 access_sale_commission_settlement_line_user,access_sale_commission_settlement_line_user,model_sale_commission_settlement_line,base.group_user,1,0,0,0
 access_sale_commission_analysis_report,access_sale_commission_analysis_report,model_sale_commission_analysis_report,base.group_user,1,0,0,0
+access_sale_order_commission_analysis_report,access_sale_order_commission_analysis_report,model_sale_order_commission_analysis_report,base.group_user,1,0,0,0
```
pedrobaeza: Shouldn't be sale user?

sale_commission/report/sale_commission_analysis_report.py
eLBati: done

sale_commission/report/sale_commission_analysis_report_view.xml
eLBati: done

sale_commission/security/ir.model.access.csv
eLBati: Modified to `sales_team.group_sale_salesman`

mrcast:
@eLBati :
- create a customer
- create an agent
- associate agent on customer
- create a sale order with a line with e new product
- saving, I obtain the warning in attachment and I can't test the PR
- 
image

mrcast:
@eLBati : Sorry, it was my mistake, don't consider my previous message

mrcast (approved):
Functional test, ok

pedrobaeza:
/ocabot merge major

Now it needs to be ported to v13

PR #280, opened 2021-02-09
[12.0] IMP sale_commission allowing to filter sale orders and invoices by agents

eLBati:


sale_commission/models/account_invoice.py
```diff
@@ -19,6 +19,20 @@ def _compute_commission_total(self):
...
+    @api.multi
+    def _compute_agents(self):
+        for so in self:
+            so.agent_ids = [(6, 0, so.mapped("invoice_line_ids.agents.agent").ids)]
+
+    @api.model
+    def _search_agents(self, operator, value):
+        sol_agents = self.env["sale.order.line.agent"].search(
+            [("agent", operator, value)])
+        return [('id', 'in', sol_agents.object_id.order_id.ids)]
```
pedrobaeza: This is going to fail:
```suggestion
        return [('id', 'in', sol_agents.mapped("object_id.order_id").ids)]
```

pedrobaeza (changes requested):
Please add tests to cover the new code.

sale_commission/models/account_invoice.py
eLBati: oops, fixed

eLBati:
@pedrobaeza tests added

pedrobaeza (approved):

mrcast (approved):
Functional test ok!

pedrobaeza:
Also interesting for v13.

marcelsavegnago (approved):

PR #282, opened 2021-02-10
[13.0][FIX] Context overriden from invoice_line_ids

ao-landoo:
We've found two problems which take place when an invoice is created.

- When adding an invoice lines, products that can be purchased are shown (if only purchase_ok is ticked, product is not showing up).

- If the customer has a different language from the default and the product being added has a description in the customer's language, the description is not displayed in the customer's language (example: default language is spanish, but customer has english and the product also has the description in english).

The solution has been keeping the core context of the invoice_lines + adding the partner_id which this module adds.

@pedrobaeza can you review this?

### DIFF
--- a/sale_commission/models/account_move.py
+++ b/sale_commission/models/account_move.py
@@ -2,6 +2,8 @@
 # Copyright 2020 Tecnativa - Manuel Calero
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from lxml import etree
+
 from odoo import _, api, exceptions, fields, models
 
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please check Travis. This is still not very inheritable, as this is overriding other possible values. You should do it through `fields_view_get` overriding, like this one: https://github.com/OCA/sale-workflow/blob/55356f0c96861490a9dcf6186d295f7d113e33ee/sale_order_general_discount/models/sale_order.py#L25

ao-landoo:
> Please check Travis. This is still not very inheritable, as this is overriding other possible values. You should do it through `fields_view_get` overriding, like this one: https://github.com/OCA/sale-workflow/blob/55356f0c96861490a9dcf6186d295f7d113e33ee/sale_order_general_discount/models/sale_order.py#L25

It should be ready now @pedrobaeza

sale_commission/models/account_move.py
```diff
@@ -36,6 +38,25 @@ def post(self):
 
     def recompute_lines_agents(self):
         self.mapped("invoice_line_ids").recompute_agents()
+        
+    @api.model
+    def fields_view_get(
+        self, view_id=None, view_type="form", toolbar=False, submenu=False
+    ):
+        res = super(AccountMove, self).fields_view_get(
```
pedrobaeza, line 49: Add a docstring here commenting why doing this way:
```suggestion
        """Inject in this method the needed context for not removing other
        possible context values.
        """
        res = super(AccountMove, self).fields_view_get(
```

pedrobaeza (changes requested):
Please check pre-commit, that is failing. You can install locally and run it for avoiding this problem. If you want, I can perform the change for not delaying this more time.

ao-landoo:
> Please check pre-commit, that is failing. You can install locally and run it for avoiding this problem. If you want, I can perform the change for not delaying this more time.

If you could perform it I would really appreciate it.

ao-landoo:
@pedrobaeza It's asking for approval

pedrobaeza (approved):
Thanks for the changes, and sorry for not doing the pre-commit stuff, as I didn't have time for it.

PR #285, opened 2021-02-12
[13.0][MIG] hr_commission: Migration to 13.0

joao-p-marques:
@Tecnativa
TT26922

ping @pedrobaeza @victoralmau


hr_commission/views/res_partner_view.xml
```diff
@@ -0,0 +1,17 @@
...
+
+    <record model="ir.ui.view" id="view_partner_form_agent">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="sale_commission.view_partner_form_agent" />
+        <field name="arch" type="xml">
+            <field name="commission_id" position="after">
+                <field
+                    name="employee"
+                    attrs="{'invisible': [('agent_type', '!=', 'salesman')]}"
+                />
```
joao-p-marques, line 12: I think we could optimize this field as we are adding a relation from ResPartner to HrEmployee here. This field is supposed to indicate whether the Partner is an employee or not. As we have that data from the ResPartner -> ResUser -> HrEmployeeRelation, we could do something like
```python
    @api.depends("agent_type")
    @api.onchange("employee_id")
    def _onchange_employee_id(self):
        if self.employee_id and self.agent_type == "salesman":
            self.employee = True
```
kudos @victoralmau 
@pedrobaeza WDYT?

victoralmau: IMO it's a good idea (remove partner view and define employee value with onchange method).

joao-p-marques: Maybe the partner view can be left as and indicative that said partner is an employee, but this improvement would help make it more logical IMHO.

pedrobaeza: You are not setting an employee, but computing it, checking if there's a user linked to this partner, and such user has a linked employee. We can make it computed writable though, for allowing to link employees without user, but in that case, we must prevent to put the same employee in several agents.

joao-p-marques: Ok, would it be something like this, then?
```python
    employee = fields.Boolean(compute="_compute_employee", inverse="_set_employee")

    @api.depends("agent_type", "employee_id")
    def _compute_employee(self):
        for record in self:
            if record.employee_id and record.agent_type == "salesman":
                record.employee = True
            else:
                record.employee = False

    def _set_employee(self):
        for record in self:
            record.employee = record.employee
```

pedrobaeza: No, a computed writable is a field with `compute="x"`, `store=True` and `readonly=False`. It's a new feature of v13 framework. You can see this video with a workshop I gave for OCA days: https://www.youtube.com/watch?v=lE4e4G45BHo

joao-p-marques: Ok, thanks. Would https://github.com/OCA/commission/pull/285/commits/f8cf27bfdeca4abe51e0205358529df8120d2adf work, then?

pedrobaeza: I'm afraid not. I was talking about `employee_id` field itself.

joao-p-marques: Ok, thanks again. In the latest commit https://github.com/OCA/commission/pull/285/commits/2e57ffe191ceb0963f9ffd20d630a21f782a0e25 I convert `employee_id` to a computed writable, as you said, and also apply the dynamic update to the `employee` field. This is a Boolean field, not really indicating the relation, but still IMHO it makes sense given the relation introduced with this module.

pedrobaeza: But why do you need such boolean field?

joao-p-marques: The field comes from Odoo core, and AFAICS it only exits to indicate if said partner is an employee or not, by manual edit. As we introduce this relation here, I think it makes sense to automatically set it to true if the relation exists, while still allowing for manual edit.

pedrobaeza: OK, I see. Let's keep it, yes, although you need to provide the compatibility layer in case other module does the same with this field. See https://github.com/OCA/odoo-community.org/pull/50/files#diff-8d1522a51d35f4f54e73c7645bb78aff2d5a74e7a991701a6c3b50bb0fa76e29R838

pedrobaeza:
Check also Travis status, and please document the overwriting of the field.

joao-p-marques:
@pedrobaeza the latest commit should fix the tests :+1:

pedrobaeza (approved):

PR #287, opened 2021-02-17
[13.0][IMP] sale_commission: Add optional in commission_status field related to order view

victoralmau:
Add optional in `commission_status` field related to order view

Please @joao-p-marques and @sergio-teruel can you review it?.

@Tecnativa TT28323

### DIFF
--- a/sale_commission/views/sale_order_view.xml
+++ b/sale_commission/views/sale_order_view.xml
@@ -17,7 +17,7 @@
                     attrs="{'invisible': ['|', ('commission_free', '=', True), ('invoice_lines', '!=', [])]}"
                     type="object"
                 />
-                <field name="commission_status" />
+                <field name="commission_status" optional="show" />
             </xpath>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can buttons be hidden as well?

victoralmau:
Is not possible to apply "_optional_" to buttons.
IMO maybe it's a good idea to show buttons only to "Sale Admin" group users.
What do you think about it?

pedrobaeza:
Well, it's a big functional change. Till now, any salesman can modify the agents (but not the amounts). And anyway, they can't see the detail, so in any case, the button must be there (maybe hidden), but the popup to be editable or not.

victoralmau:
Ping @joao-p-marques and @sergio-teruel

sergio-teruel (approved):

joao-p-marques (approved):

carlosdauden (approved):

PR #288, opened 2021-02-20
12.0 mig sale commission product new

scigghia:
Reworked sale commission product


josepmy:
@scigghia From a existing draft invoice if you add a new line and try to save appear an error:

`Operation cannot be completed:
Create/Update: missing mandatory field.

Model account.invoice.line.agent (account.invoice.line.agent), Field: Agent (agent)`

Seems that the Agent ID is missing.

Captura de pantalla 2021-09-06 a las 8 44 36

PR #290, opened 2021-02-26
[13.0][FIX] sale_commission: Invoice lines default values

ioans73:
We have detected that when creating an invoice manually, some fields are not filled in the lines by default, such as the partner_id field, which are filled in without having this module installed and this is causing us other problems (this happens because this module modifies the context of the invoice_line_ids field).

We understand that for previous versions there was no problem, since the invoices and accounting entries were separated and when an invoice was validated and the entry generated, they were filled in automatically.


pedrobaeza:
Duplicated of #282

PR #292, opened 2021-03-17
[12.0][FIX] sale_commission

SilvioGregorini:
Fixed bug that prevented the constrain to work properly (could not access field directly from `Many2many` field).

### DIFF
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -189,7 +189,6 @@ class SettlementLine(models.Model):
     @api.constrains('settlement', 'agent_line')
     def _check_company(self):
         for record in self:
-            if record.agent_line.company_id != record.company_id:
-                raise UserError(_(
-                    'Company must be the same'
-                ))
+            for line in record.agent_line:
+                if line.company_id != record.company_id:
+                    raise UserError(_("Company must be the same"))


matteoopenf (approved):

sale_commission/models/settlement.py
```diff
@@ -189,7 +189,6 @@ class SettlementLine(models.Model):
     @api.constrains('settlement', 'agent_line')
     def _check_company(self):
         for record in self:
-            if record.agent_line.company_id != record.company_id:
-                raise UserError(_(
-                    'Company must be the same'
-                ))
+            for line in record.agent_line:
+                if line.company_id != record.company_id:
+                    raise UserError(_("Company must be the same"))
```
pedrobaeza, line 194:
```suggestion
       for line in self.mapped("agent_line"):
            if line.company_id != record.company_id:
                raise UserError(_("Company must be the same"))
```
Isn't that a single value anyway?

SilvioGregorini: @pedrobaeza `self` is potentially a recordset in which every record is linked to a different company. Thus, it wouldn't make much sense to use `self.mapped("agent_line")`: in that way, we will retrieve every line disregarding the settlement line they belong to (instead, we need to check whether the agent line's company matches its own parent settlement line's company).
Moreover, if you simply use
```
    @api.constrains('settlement', 'agent_line')
    def _check_company(self):
        for line in self.mapped("agent_line"):
            if line.company_id != record.company_id:
                raise UserError(_("Company must be the same"))

```
without cycling `self`, Python would become quite mad because variable `record` would not be defined :smile:

pedrobaeza: OK, sorry, you are right that we need to know the intermediate record. Do you have any real case where the singleton error raised?

SilvioGregorini: @pedrobaeza it happened when we tried to create a custom wizard (for a customer of ours) where we need to create new settlement lines linked to multiple agent lines at once, a simple feature that we tried to obtain with a trivial 
```
self.env['sale.commission.settlement.line'].create({..., 'agent_line': [(6, 0, ids)]}
```
Unfortunately, the constrain failed if `ids` was an empty list which led to an error being raised (because the current behaviour doesn't distinguish whether there's at least 1 agent line or not, so `record.agent_line.company_id` is always an empty recordset `res.company()` if no line is set), and crashed Odoo when `ids` was a list with 2 or more elements (because of the singleton error).

pedrobaeza: OK, don't worry, it was just curiosity, as I didn't find such case with OCA modules, but not being totally correct, it's OK to fix it here in the base module. Maybe your customization also fits here in OCA as extension module :wink:

pedrobaeza (approved):

PR #296, opened 2021-04-14
[FIX] sale_commission_formula: test

chienandalu:
The invoice post() was done twice, which is unnecessary and with the
recent upstream changes raises an exception

cc @Tecnativa TT29179


pedrobaeza (approved):

pedrobaeza:
Fast tracking for solving the test error:

/ocabot merge patch

PR #299, opened 2021-04-23
[13.0][FIX] Commission fields should required only if partner is agent

renatonlima:
When some module creates one partners or load by XML file, the fields agent_type and settlement are null, if you edit in interface this partner and it is not a sale agent you can't save because these fields required.

I added the required using "attrs" attribute in these fields.

### DIFF
--- a/sale_commission/views/res_partner_view.xml
+++ b/sale_commission/views/res_partner_view.xml
@@ -32,14 +32,20 @@
                 >
                     <group>
                         <group>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please check pre-commit

pedrobaeza (approved):

rvalyi (approved):

marcelsavegnago (approved):

PR #273, opened 2020-12-17
[14.0][MIG] sale_commission

mchristoph:

PR #302, opened 2021-05-25
[14.0][MIG] sale_commission: Migration to 14.0

fredzamoabg:
sale_commission module migration from 13.0
(alignment of https://github.com/OCA/commission/pull/273)


rousseldenis (approved):
Code review

sale_commission/wizard/wizard_settle.py
```diff
@@ -114,6 +114,7 @@ def action_settle(self):
                                 )
                             )
                         settlement_ids.append(settlement.id)
+                    line.read(["amount"])  # force amount computation
```
pedrobaeza: This seems weird and shouldn't be needed. What are the steps for needing this?

fredzamoabg: You are right, but running tests was found error related amount computation that causing exception per validation in **_check_settle_integrity** method. I'll try to find out the reason in order to remove that line.

thanks

fredzamoabg: removed and added amount field computation in tests

pedrobaeza (approved):

PR #308, opened 2021-06-29
[12.0] FIX sale_commission preventing to cancel invoice with settled commission lines.

eLBati:
Steps to reproduce:
    
     - Create an agent
     - Create and validate an invoice setting that agent
     - Settle the commission
     - Cancel the invoice
    
    Observed: the invoice is canceled
    
    Desired: cancellation is blocked

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -35,7 +35,12 @@ class AccountInvoice(models.Model):
         return [('id', 'in', ail_agents.mapped("object_id.invoice_id").ids)]
 
     def action_cancel(self):
-        """Put settlements associated to the invoices in exception."""
+        """Put settlements associated to the invoices in exception
+        and check settled lines"""
+        if any(self.mapped("invoice_line_ids.any_settled")):
+            raise exceptions.ValidationError(
+                _("You can't cancel an invoice with settled lines"),
+            )
         settlements = self.env['sale.commission.settlement'].search(
             [('invoice', 'in', self.ids)])
         settlements.write({'state': 'except_invoice'})


pedrobaeza:
I think the problem is that the invoice shouldn't be cancelled in first place, so you need to extend the constraint variables to state as well and check if the target state is different from ('open', 'paid').

eLBati:
@pedrobaeza thanks.

Trying with
```
diff --git a/sale_commission/models/account_invoice.py b/sale_commission/models/account_invoice.py
index fbb0e1c..3f38eb2 100644
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -185,7 +185,7 @@ class AccountInvoiceLineAgent(models.Model):
         for line in self:
             line.company_id = line.object_id.company_id
 
-    @api.constrains('agent', 'amount')
+    @api.constrains('agent', 'amount', 'invoice.state')
     def _check_settle_integrity(self):
         for record in self:
             if any(record.mapped('settled')):
```
`_check_settle_integrity` is not executed when canceling the invoice.

So, updated the PR modifying `action_cancel`

pedrobaeza (commented):
Now that I see the code, the idea is to have the settlement in `except_invoice` state, but not to prevent the cancel, so maybe it's not good idea to block them. Sorry for not remembering that detail.

Which flow do you think is more acceptable? Allow the invoice to be re-validate it in that case? Forbid at all the cancellation?

mrcast (approved):
Functional test ok!

eLBati:
@pedrobaeza from what I can see, the settlement becomes `except_invoice` when the agent's invoice is canceled, not the customer's one

pedrobaeza (approved):
OK, I see!

pedrobaeza:
It would be great to fw-port this to v13/v14

eLBati:
Here they are
https://github.com/OCA/commission/pull/309
https://github.com/OCA/commission/pull/310

PR #310, opened 2021-07-01
[14.0] FIX sale_commission preventing to cancel invoice with settled commiss…

eLBati:
…ion lines.

Steps to reproduce:

 - Create an agent
 - Create and validate an invoice setting that agent
 - Settle the commission
 - Cancel the invoice

Observed: the invoice is canceled

Desired: cancellation is blocked


pedrobaeza (approved):

etobella (approved):
Makes sense

PR #312, opened 2021-07-12
[IMP] sale_commission: Allow to group invoices

etobella:
With 13.0 version, it was lost the option to merge all the settlements of a partner and company to a single invoice. With this change, you can select the behavior that you will use

### DIFF
--- a/sale_commission/__manifest__.py
+++ b/sale_commission/__manifest__.py
@@ -2,7 +2,7 @@
 # Copyright 2020 Tecnativa - Manuel Calero
 {
     "name": "Sales commissions",
-    "version": "13.0.1.5.2",
+    "version": "13.0.1.6.0",
     "author": "Tecnativa," "Odoo Community Association (OCA)",
     "category": "Sales Management",
     "license": "AGPL-3",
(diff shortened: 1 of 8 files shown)


pedrobaeza:
More than "merge invoices", it's to generate grouped settlements invoices...

Where this option was in 12.0?

etobella:
You are right about the name.

There was not an option for that, however, a hook existed that allowed to make the change:

https://github.com/OCA/commission/blob/12.0/sale_commission/models/settlement.py#L136

This PR has copied the logic from odoo `sale` module.

pedrobaeza:
2 comments:

- Check the recent bug we discovered on sale invoicing: https://github.com/odoo/odoo/pull/71686
- Maybe you can split again the logic of header/lines (passing `Form` as argument), for improving both the readability and the extensibility.

etobella:
Thanks for the fix @pedrobaeza 
I would prefer to not use the Form as it should be slower than this method

pedrobaeza:
> I would prefer to not use the Form as it should be slower than this method

That's not true, as now, if you have N invoices, N Forms are instantiated and N lines, and then converted to vals, and then process the vals. With my proposal, only 1 Form will be instantiated and N lines, so it's worth to do it.

etobella:
True, I didn't notice that. I will check, thanks!!!

etobella:
@pedrobaeza Comments attended, WDYT?

pedrobaeza (approved):

albariera (approved):
LGTM :+1:

pedrobaeza:
Pleas forward-port it to v14

PR #315, opened 2021-07-23
[14.0][FIX] sale_commission

max3903:


bodedra (approved):
LG :+1:

RLeeGML (approved):
LGTM

pedrobaeza:
Maybe someone would one to have an extra option for deciding if the `in_payment` state is considered or not, but for now, let's go with this:

/ocabot merge patch

PR #316, opened 2021-07-26
[14.0][ADD] sale_commission_product

JoanSForgeFlow:
New module that allows to select an agent fora product.
When this product is sold, the agent will be added in the sale order line automatically.

cc @ForgeFlow


pedrobaeza:
There's already #288

PR #318, opened 2021-08-23
[12.0] IMP sale_commission sale_commission allowing to add agents only by view_sale_commission_mixin_agent_only, aligning to 13.0

eLBati:
Also fix the following use case:

 - Create a quotation template with a quotation line
 - Create a quotation selecting the new template > the line will be added
 - Open line form and try to add an agent

Get TypeError: fieldsInfo is undefined

### DIFF
--- a/sale_commission/views/sale_order_view.xml
+++ b/sale_commission/views/sale_order_view.xml
@@ -19,8 +19,6 @@
             </xpath>
             <xpath expr="//field[@name='order_line']/form//field[@name='customer_lead']/.." position="after">
                 <field name="commission_free"/>
-                <field name="agents"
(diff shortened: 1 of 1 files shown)


mrcast (approved):
Functional test ok!

matteoopenf (approved):

pedrobaeza (changes requested):
I have aborted the merge, as this PR will provoke the problem of not saving correctly the agents, and that's why it's put on the view. Maybe the solution is to use `force_save=1`?

chienandalu (approved):

pedrobaeza (approved):
Hit the problem also, so the fix seems legit, and the write override performs the assignation on record save. Although not optimal, but works well enough.

PR #319, opened 2021-09-16
[FIX] migration scripts for 13.0

etobella:
I found some errors when migrating to 13.0 using OpenUpgrade


etobella:
@pedrobaeza

sale_commission_formula/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -1,13 +0,0 @@
-# Copyright 2020 NextERP Romania SRL
```
pedrobaeza: You have forgotten to add the post-migration file.

sale_commission/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -8,3 +8,36 @@ def migrate(env, version):
...
         env.cr, "sale_commission", "migrations/13.0.1.0.0/noupdate_changes.xml"
     )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_invoice_line_agent aila
+        SET object_id = aml.id
+        FROM account_move_line aml
+        WHERE aml.old_invoice_line_id = aila.{}
+    """.format(
```
pedrobaeza: Please use `sql.SQL` and `sql.Identifier` for proper SQL build.

pedrobaeza (changes requested):

sale_commission/migrations/13.0.1.0.0/post-migration.py
etobella: Sure!

etobella:
@pedrobaeza Comments attended :wink:

pedrobaeza (approved):

PR #320, opened 2021-09-23
[13.0][OU-IMP] sale_commission: migration: create column in order to abort compute

etobella:
Without it, it is computing field for each line. We are already doing it on the post-migration


pedrobaeza (approved):

pedrobaeza:
You can put it as `[OU-IMP]` for knowing that it's something related to OpenUpgrade. Merging directly as ocabot doesn't add value here.

PR #321, opened 2021-09-24
[12.0] ADD sale_commission_settlement_partial_invoice

eLBati:
Allow to generate partial agent invoice from settlement


mrcast (commented):
@eLBati : Settlement just created, are immediately set as invoiced
**STEP**
- create an invoice with an agent
- lunch "settlements commision"
- the created settlement is in state invoiced instead of state "Settled"
Schermata del 2021-09-24 18-08-29

eLBati:
@mrcast thanks, fixed

mrcast (approved):
Functional test, ok!

eLBati:
I think this could be merged as `beta`

PR #323, opened 2021-09-28
[13.0][OU-IMP] sale_commmission: compute fields using SQL

etobella:
I found other fields that took long time to compute. Now it should be faster. On my case, from 17 minutes to 3 minutes

### DIFF
--- a/sale_commission/migrations/13.0.1.0.0/post-migration.py
+++ b/sale_commission/migrations/13.0.1.0.0/post-migration.py
@@ -33,6 +33,15 @@ def migrate(env, version):
     )
     openupgrade.logged_query(
         env.cr,
+        """
+        UPDATE sale_commission_settlement scs
+        SET invoice_id = am.id
+        FROM account_move am
+        WHERE am.old_invoice_id = scs.invoice
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
         sql.SQL(
             """
         UPDATE sale_commission_settlement_line scsl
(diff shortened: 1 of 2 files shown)


sale_commission/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -42,3 +51,24 @@ def migrate(env, version):
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE account_move_line aml
+            SET commission_free = pt.commission_free
+            FROM product_product pp
+                INNER JOIN product_template pt ON pt.id = pp.product_tmpl_id
+            WHERE aml.product_id = pp.id AND pt.commission_free = TRUE
```
pedrobaeza: Just FYI, this is the same:

```suggestion
            WHERE aml.product_id = pp.id AND pt.commission_free
```

pedrobaeza: I don't think you should do this, but take it from the source account.invoice.line, as the check may vary over time, having different values in the resulting documents.

pedrobaeza (changes requested):

sale_commission/migrations/13.0.1.0.0/post-migration.py
etobella: Sure, thanks!!! I didn't notice the information could be found in invoice line. It has more sense to import data from there in order to keep the same information.

sale_commission/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -42,3 +51,23 @@ def migrate(env, version):
...
+                AND ai.commission_total != 0.0
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE account_move_line aml
+            SET commission_free = ail.commission_free
+            FROM account_invoice_line ail
+            WHERE aml.old_invoice_line_id = ail.id AND ail.commission_free
```
pedrobaeza, line 71: I'm wondering if this field can contain nulls by default and then you need to assign all values

etobella: I think we shouldn't reassign the value, as this value comes from the original invoice

pedrobaeza (approved):

PR #324, opened 2021-10-01
[14.0][IMP] sale_commission: Link created invoice

hildickethan:
- Change to `invoice_id` field in `sale.commission.settlement` to link to the invoice created to pay the agent's commissions. 

Currently this field is just taking the first invoice that is being used in it's commissions, which is somewhat redundant information, hence it's TODO to be removed.

I think changing it's use to this is an appropriate way to fulfill the TODO.

This change also doesn't affect current records, as it's previous compute was stored, so this will only affect future settlement creations. A migration to retroactively fix this may be considered needed, but since that's old accounting that will have been closed, I don't think it's strictly necessary to change.

### DIFF
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -45,12 +45,7 @@ class Settlement(models.Model):
         string="Generated invoice",
         readonly=True,
     )
-    # TODO: To be removed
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think we should create another field for such purpose.

PR #330, opened 2021-11-21
[FIX] sale_commission: Fix grouped invoices creation

etobella:
When using grouped invoices, an error could be created because sign was changed with type. Now it should work properly, as the sign is changed at the end.

For example, if you had two settlements, one of 10€ and one of -10€, the old result was an invoice of 20€. Now, the result is an invoice of 0€.

### DIFF
--- a/sale_commission/models/account_move.py
+++ b/sale_commission/models/account_move.py
@@ -93,6 +93,17 @@ class AccountMoveLine(models.Model):
                     record.move_id.partner_id
                 )
 
+    def _copy_data_extend_business_fields(self, values):
+        """
+        We don't want to loose the settlement from the line when reversing the line if
+        it was a refund.
+        We need to include it, but as we don't want change it everytime, we will add
+        the data when a context key is passed
+        """
+        super()._copy_data_extend_business_fields(values)
+        if self.settlement_id and self.env.context.get("include_settlement", False):
+            values["settlement_id"] = self.settlement_id.id
+
 
 class AccountInvoiceLineAgent(models.Model):
     _inherit = "sale.commission.line.mixin"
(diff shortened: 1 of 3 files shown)


sale_commission/models/settlement.py
```diff
@@ -106,7 +107,7 @@ def _prepare_invoice(self, journal, product, date=False):
         for settlement in self:
             with move_form.invoice_line_ids.new() as line_form:
                 line_form.product_id = product
-                line_form.quantity = 1
+                line_form.quantity = -1 if settlement.total < 0 else 1
```
pedrobaeza, line 110: But we don't want negative invoices...

etobella: I know, at the end it is transformed to a refund. Similary to sale

sale_commission/models/account_move.py
```diff
@@ -93,6 +93,11 @@ def _compute_agent_ids(self):
                     record.move_id.partner_id
                 )
 
+    def _copy_data_extend_business_fields(self, values):
+        super()._copy_data_extend_business_fields(values)
+        if self.settlement_id and self.env.context.get("include_settlement", False):
```
pedrobaeza, line 104: And this?

pedrobaeza:
Maybe you can include a test for avoiding regressions in this.

sale_commission/models/settlement.py
pedrobaeza: OK, seen.

sale_commission/models/account_move.py
etobella: Without this you loose the settlement_line_id. The only way I found without breaking the old logic was this.

etobella:
@pedrobaeza comments attended

sale_commission/models/account_move.py
pedrobaeza: Then please put a docstring specifying why this is needed and from where this is called. If not, later on the future we won't remember why is this here, and we will remove it.

pedrobaeza (approved):

PR #341, opened 2022-03-02
[14.0][IMP] sale_commission: add new commission type "type"

ilyasProgrammer:
This PR adds a new type to "commission type" model allowing user to structure commission type similarly to how pricelist is structured. This means: 

- commission type lines are now available, where commission rate can be set on global/category/product/variant level. Lines are sorted from variant level to global level, just like in pricelist.

- commission rate can be either a percentage or a fixed value

- in SO line, the first applicable commission type line is applied.

- there is an extra right "Display full agent details on sale order line" in user to enhance agent commission details on SO line and check what commission type line has been applied.


francesco-ooops (approved):
LGTM

pedrobaeza (changes requested):
This shouldn't touch base module adding such extra models and things that increase a lot the maintenance burden and even doesn't have a migration plan

francesco-ooops:
> This shouldn't touch base module adding such extra models and things that increase a lot the maintenance burden and even doesn't have a migration plan

Ok, I tried sending an email about this module to crm-sales-marketing mailing list requesting feedback but didn't receive any answer. We'll gladly make changes and adapt code in a new module but it would be good if there was an outlet to discuss implementation before all the cleanup work and test coverage is done. 

Where do you suggest future developments could be talked in advance?

pedrobaeza:
In an issue in the repo, but you should always have to take into account that heavy refactorings in stable branches are not advisable, and also that you can't make a lot more complex the base module for a special case.

PR #344, opened 2022-03-08
[12.0][FIX] set settlements state to except_invoice when unlink commission invoice

renatonlima:
Today if you create a draft commission invoices and cancel, all settlements related is changed to Except Invoice state. but it you unlink a draft commission invoices,  all settlements related still with invoced state.

After review and approval I can port this to 13.0 and 14.0 branches.


rvalyi (approved):
LGTM. I guess few people will actually delete invoices and this is probably why this issue has never been discussed. But I agree it's more consistent this way.

renatonlima:
Hi guys, To keep module quality I will add a test for unlink commission invoice

PR #343, opened 2022-03-08
[15.0][MIG] Commission : Migration to 15.0

manuelcalerosolis:
This is the migration of the "commissions" repo to 15.0 but there are important changes with respect to previous versions.

PR #345, opened 2022-03-10
15.0 mig sale commission

slominchar:


pedrobaeza:
This is being handled partially in #343. See discussion in https://github.com/OCA/commission/issues/328#issuecomment-986704280

PR #346, opened 2022-03-11
[12.0][REF] The way the commission_free is taken

antoniospneto:
This PR changes the way the "commission_free" is taken when calculating the amount, instead of taking the product, it is taken directly from the invoice_line (object_id) , which has the commission_free as a related to the product's commission_free.

This will not change any behavior, but it makes it easier for those who inherit the module, we are making a module for Brazil where the commission_free becomes a computed and it will not only depend on the product, but also on other information.

### DIFF
--- a/sale_commission/models/account_invoice.py
+++ b/sale_commission/models/account_invoice.py
@@ -164,7 +164,7 @@ class AccountInvoiceLineAgent(models.Model):
         readonly=True,
     )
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Using the related field makes that the computations are not done until the record is saved, and that's why this is not done such way.

felipemotter (approved):

PR #349, opened 2022-03-21
[ADD] sale_commission_supplierinfo

olgamarcocb:
Adding a new type of commission (supplierinfo) which calculates the result of the agent product price and the product quantity.

### DIFF
--- a/sale_commission_supplierinfo/models/sale_commission.py
+++ b/sale_commission_supplierinfo/models/sale_commission.py
@@ -1,0 +1,11 @@
+# Copyright 2022 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import fields, models
+
+
+class SaleCommission(models.Model):
+    _inherit = "sale.commission"
+
+    commission_type = fields.Selection(selection_add=[("supplierinfo", "Supplierinfo")])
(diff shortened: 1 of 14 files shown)


sale_commission_supplierinfo/models/sale_commission.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+from odoo import fields, models
+
+
+class SaleCommission(models.Model):
+    _inherit = "sale.commission"
+
+    commission_type = fields.Selection(selection_add=[("supplierinfo", "Supplierinfo")])
+    supplierinfo = fields.Text(
```
etobella: Do you have this field for any reason?

olgamarcocb: Ups, solved!

sale_commission_supplierinfo/models/sale_commission_mixin.py
```diff
@@ -0,0 +1,44 @@
...
+        """Get the commission amount for the data given. To be called by
+        compute methods of children models.
+        """
+        self.ensure_one()
+        if (
+            not product.commission_free
+            and commission
+            and commission.commission_type == "supplierinfo"
+        ):
+            params = {"order_id": self.object_id}
```
etobella: Mmmh, this might be a problem, because, it is not an order (it might be a sale.order.line or an account.move.line)

olgamarcocb: You're right

PR #353, opened 2022-04-14
[13.0][IMP] sale_commission: New bi-weekly settlement period

pedrobaeza:
There will be 2 settlement per month, one covering the first 15 days, and the other for the rest of the month.

@Tecnativa TT35385

PR #358, opened 2022-05-10
Forward port missing commits. Forwarding all missing commits from 13.0 to 14.0

olgamarcocb:
Includes:
- #311
- #312
- #330
- #353

@etobella @pedrobaeza

### DIFF
--- a/sale_commission/tests/test_sale_commission.py
+++ b/sale_commission/tests/test_sale_commission.py
@@ -3,6 +3,7 @@
 # License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
 
 import dateutil.relativedelta
+from dateutil.relativedelta import relativedelta
 
 from odoo import fields
 from odoo.exceptions import UserError, ValidationError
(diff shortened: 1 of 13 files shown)


sale_commission/tests/test_sale_commission.py
```diff
@@ -171,6 +171,7 @@ def _invoice_sale_order(self, sale_order, date=None):
         if date:
             invoice.invoice_date = date
             invoice.date = date
+        invoice.flush()
```
pedrobaeza, line 175: Why do you need these `flush` if no SQL operation is done?

etobella: We need to execute all compute. commission_amount is not executed until it is needed, but it is needed once it has been settled. Then an error is raised. In order to avoid this, we need to flush.

pedrobaeza: Then put such explanation as comment here for the rest and for your future you for avoiding an accidental removal

olgamarcocb: :wink:

pedrobaeza: Thanks. Still something bugs me, as a correct dependency chain should trigger the recomputation, but I'm not going around more.

pedrobaeza (approved):

etobella (approved):

PR #362, opened 2022-06-07
[14.0][IMP] sale_commission: Link created invoice

hildickethan:
Reopen rebase of https://github.com/OCA/commission/pull/324
I will copy paste the description of PR from there

- Change to `invoice_id` field in `sale.commission.settlement` to link to the invoice created to pay the agent's commissions. 

Currently this field is just taking the first invoice that is being used in it's commissions, which is somewhat redundant information, hence it's TODO to be removed.

I think changing it's use to this is an appropriate way to fulfill the TODO.

This change also doesn't affect current records, as it's previous compute was stored, so this will only affect future settlement creations. A migration to retroactively fix this may be considered needed, but since that's old accounting that will have been closed, I don't think it's strictly necessary to change.

As commented in the PR we could have a different field, but this field seems unnecessary to begin with that I find it appropriate to replace with this, though it could likely be improved

### DIFF
--- a/sale_commission/models/settlement.py
+++ b/sale_commission/models/settlement.py
@@ -47,12 +47,7 @@ class Settlement(models.Model):
         string="Generated invoice",
         readonly=True,
     )
-    # TODO: To be removed
-    invoice_id = fields.Many2one(
-        store=True,
-        comodel_name="account.move",
-        compute="_compute_invoice_id",
-    )
+    invoice_id = fields.Many2one(comodel_name="account.move")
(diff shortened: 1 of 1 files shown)


hildickethan:
Ok I made this PR a bit blindly to fix conflict issues, but after doing some actual testing due to the Travis error it seems that what previously didn't work properly and caused the need for this change has been adjusted in some other way so it links properly. So I will close this as it isn't actually needed and the existing field actually does something now

PR #363, opened 2022-06-23
[15.0][MIG] Commission : Migration to 15.0 

manuelcalerosolis:
[This is the migration of the "commissions" repo to 15.0 but there are important changes with respect to previous versions.]

(https://github.com/OCA/commission/pull/343)


BT-aestebanez:
@manuelcalerosolis To be honest, I don't see the point on removing sale_comission in https://github.com/OCA/commission/pull/363/commits/5fb8108b80c8cb327e33f9c30e2cd4b4a134d3fb in order to add it again in https://github.com/OCA/commission/pull/363/commits/bf13df3e9a47143949c76a35e33a8149c58e1e66 This process only removes the original authors of the lines...
Imho, you can do it in two commits, the one with black, isort with only the changes done automatically, and another commit with the rest, making sure you include @dani-xtendoo as contributor.

francescapenso:
Hello any update about this PR?

yostashiro:
We are interested in working on this migration if this one is not moving forward soon.  If that happens, I was wondering whether we should create one PR covering `commission` and `sale_commission` like this one, or create two separate PRs, each covering a module.  Any preference?

pedrobaeza:
Only one, as the split occurs from the same previous module.

yostashiro (commented):
I feel that the split is not done in an optimal way although I haven't done a thorough check. For example, shouldn't settlement be part of the commission module instead of sale_commission?

There is also this comment https://github.com/OCA/commission/issues/328#issuecomment-986704280 from @pedrobaeza of intending to have `commission`, `account_commission` and `sale_commission` out of 14.0 sale_commission. I think we should go in line with this.

Edit: Sorry, commented in the wrong PR (it was meant to be here https://github.com/OCA/commission/pull/372#pullrequestreview-1109329899), but I guess it applies to this one as well...

ValentinVinagre:
Hello @manuelcalerosolis has the migration finished?

manuelcalerosolis:
Yes, it is finished, in fact I use it in several clients.
I don't know what's wrong with the PR, if you can help me I'd appreciate it.

pedrobaeza:
The problem with this one is that the refactoring is not what we talked about, so let's supersede by #374

PR #371, opened 2022-09-01
[15.0][MIG]Commission : Migration to 15.0 (split module of v14 sale_commission module)

AungKoKoLin1997:


pedrobaeza:
This is not correct, at it doesn't preserve the commit history.

PR #372, opened 2022-09-08
[15.0][MIG] commission, sale_commission : Migration to 15.0

AungKoKoLin1997:
This is the migration of sale_commission to 15.0.


yostashiro (commented):
I feel that the split is not done in an optimal way although I haven't done a thorough check. For example, shouldn't settlement be part of the commission module instead of sale_commission?

There is also this comment https://github.com/OCA/commission/issues/328#issuecomment-986704280 from @pedrobaeza of intending to have `commission`, `account_commission` and `sale_commission` out of 14.0 sale_commission. I think we should go in line with this.

commission/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "Commissions",
+    "version": "15.0.1.0.0",
+    "author": "Tecnativa," "Odoo Community Association (OCA)",
+    "category": "Sales Management",
+    "license": "AGPL-3",
+    "depends": ["product", "sale_management"],
```
yostashiro, line 7:
```suggestion
    "depends": ["product", "account"],
```

It shouldn't depend on sale_management.  Menu items should be moved under invoicing.

ValentinVinagre:
> I feel that the split is not done in an optimal way although I haven't done a thorough check. For example, shouldn't settlement be part of the commission module instead of sale_commission?
> 
> There is also this comment [#328 (comment)](https://github.com/OCA/commission/issues/328#issuecomment-986704280) from @pedrobaeza of intending to have `commission`, `account_commission` and `sale_commission` out of 14.0 sale_commission. I think we should go in line with this.

@pedrobaeza what do u think?

AungKoKoLin1997:
https://github.com/OCA/commission/pull/374

PR #376, opened 2022-10-18
[IMP][14.0] sale_commission

AshishHirapara:
Agent detail button will be visible even after the invoice is created

### DIFF
--- a/sale_commission/views/sale_order_view.xml
+++ b/sale_commission/views/sale_order_view.xml
@@ -14,7 +14,7 @@
                 <button
                     name="button_edit_agents"
                     icon="fa-users"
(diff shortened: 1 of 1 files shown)


francesco-ooops:
@pedrobaeza any opinion on this?

do you think it would be safer to add a setting in config to allow this field to be displayed after invoice is issued?

pedrobaeza (changes requested):
This is incorrect and shouldn't be allowed, and even more, it doesn't serve, as the agents have been already transferred to the invoice

francesco-ooops:
> 

Can you help me understand why user shouldn't be able to check commissions in SO after invoice is created?

pedrobaeza:
They shouldn't modify them, and as said, that information is not useful, the important one is the invoice agents information.

francesco-ooops:
@pedrobaeza ok this can be closed then, thanks

PR #385, opened 2022-11-21
[FIX] sale_commission: fix down payment invoice agents commission calculation

ilyasProgrammer:
In current version of module down payment invoice agent commission calculated wrong way.
Create Sale Order with 2 lines: Product 1 1000$, Product 2 1000$. 
Set agent commission only for 1st line 10% = 100$ for some agent.
Create down payment invoice with total 1500$.
You will get invoice agent commission 10% of 1500$ = 150$, which is higher than his commission in Sale Order (100$).

This PR makes commission calculation: 1500$ is 3/4 of 2000$. Agent commission will be 3/4 of 100$ = 75$.

### DIFF
--- a/sale_commission/models/account_move.py
+++ b/sale_commission/models/account_move.py
@@ -99,6 +99,7 @@ class AccountMoveLine(models.Model):
 
     agent_ids = fields.One2many(comodel_name="account.invoice.line.agent")
     any_settled = fields.Boolean(compute="_compute_any_settled")
+    is_downpayment = fields.Boolean(default=False)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Down payment product must be marked as commission free.

ilyasProgrammer:
@pedrobaeza What if it will be a separate module ? **sale_commission_advance**

pedrobaeza:
OK, I suppose you have the need to settle the commissions for down payments, but in my experience, they are very hard to trace, and all my customers using commissions, exclude down payments from the commission system.

PR #386, opened 2022-11-21
[ADD] sale_commission_advance: down payment invoice agents commission calculation

ilyasProgrammer:


francesco-ooops (approved):
replaces #385 

functional review ok!

@pedrobaeza could you provide an opinion? thanks!

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,84 @@
+from odoo import api, fields, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    is_downpayment = fields.Boolean(default=False, copy=False)
```
pedrobaeza: This field is conflicting: on one part, the generic name can collapse with another and it's only a helper one. On the other, it won't be set for down payments done prior to the module installation.

You should deduce if something is a down payment other way.

sale_commission_advance/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+* Create Sale order with agents commissions.
+* Create down payment invoice. Make sure down payment product is not commission free.
+* Open created invoice. You will see proportional agents commissions calculated.
```
francesco-ooops, line 3:
```suggestion
* Open created invoice. You will see proportional agents commissions calculated.
* When creating regular invoice with deducted down payment, also agents commissions from down payment will be deducted.
```

francesco-ooops (approved):
LGTM

@pedrobaeza can you check it again? thank you :)

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+            rec.recompute_lines_agents()
+        return rec
+
+    def has_related_sale_with_down_payment(self):
+        if any(
+            self.invoice_line_ids.mapped(
+                lambda x: x.sale_line_ids and x.sale_line_ids.is_downpayment
+            )
+        ):
+            return True
```
pedrobaeza:
```suggestion
        return any(
            self.invoice_line_ids.mapped(
                lambda x: x.sale_line_ids and x.sale_line_ids.is_downpayment
            )
        )
```

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    @api.model
+    def create(self, vals):
+        rec = super(AccountMove, self).create(vals)
```
pedrobaeza: And please add a docstring why you need to do this.

```suggestion
        rec = super().create(vals)
```

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    @api.depends("move_id.partner_id")
+    def _compute_agent_ids(self):
+        self.agent_ids = False  # for resetting previous agents
+        for record in self.filtered(
+            lambda x: x.move_id.partner_id and x.move_id.move_type[:3] == "out"
+        ):
+            if record.sale_line_ids and record.sale_line_ids.is_downpayment:
```
pedrobaeza: No need of the double if, as if no sale line, it will be False as well.

```suggestion
            if record.sale_line_ids.is_downpayment:
```

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+                # Down payment invoice
+                agent_ids = record._prepare_agents_vals_down_payment()
+                record.update({"agent_ids": agent_ids})
+            else:
+                # Regular invoice
```
pedrobaeza: You should call super here (or better, extract the subset and call super only once).

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+                if not record.commission_free and record.product_id:
+                    record.agent_ids = record._prepare_agents_vals_partner(
+                        record.move_id.partner_id
+                    )
+
+    def _prepare_agents_vals_partner(self, partner):
+        if not self.move_id.has_related_sale_with_down_payment():
+            return super(AccountMoveLine, self)._prepare_agents_vals_partner(partner)
```
pedrobaeza:
```suggestion
            return super()._prepare_agents_vals_partner(partner)
```

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+        self.sale_line_ids.ensure_one()
```
pedrobaeza: Why this and after perform a filtered?

pedrobaeza (commented):
It's very convenient to add tests to cover this module for being sure everything works as expected, both for a downpayment case and the normal one with this module installed.

ilyasProgrammer:
@pedrobaeza Updated

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    @api.model
+    def create(self, vals):
+        rec = super().create(vals)
+        if rec.has_related_sale_with_down_payment():
+            # to get correct commission amount
```
pedrobaeza: This is the what, not the why it doesn't have the correct commission amount. It's also interesting to be put as docstring, not inline comment, as the overwrite of the method is just for this reason.

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+            rec.recompute_lines_agents()
+        return rec
+
+    def has_related_sale_with_down_payment(self):
+        return any(
+            self.invoice_line_ids.mapped(
+                lambda x: x.sale_line_ids and x.sale_line_ids.is_downpayment
+            )
+        )
```
pedrobaeza: This can be simplified this way:

```suggestion
        return any(self.invoice_line_ids.sale_line_ids.is_downpayment)
```

```
>>> any([])
False
```

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+        self.agent_ids = False  # for resetting previous agents
+        down_payment_items = self.filtered(
+            lambda x: x.move_id.partner_id
+            and x.move_id.move_type[:3] == "out"
+            and x.sale_line_ids.is_downpayment
+        )
+        for record in down_payment_items:
+            agent_ids = record._prepare_agents_vals_down_payment()
+            record.update({"agent_ids": agent_ids})
+        regular_items = self.filtered(
```
pedrobaeza: Why computing twice?

```suggestion
        regular_items = self - down_payment_items
```

sale_commission_advance/models/account_move.py
```diff
@@ -0,0 +1,129 @@
...
+                    },
+                )
+            )
+        return res
+
+
+class AccountInvoiceLineAgent(models.Model):
+    _inherit = "account.invoice.line.agent"
+
+    commission_id = fields.Many2one(
```
pedrobaeza: Why redefining the field?

sale_commission_advance/models/sale_order.py
```diff
@@ -0,0 +1,44 @@
...
+        for record in self.filtered(lambda x: x.order_id.partner_id):
+            if record.is_downpayment:
+                # Down payment SOL
+                agent_ids = record._prepare_agents_vals_partner_down_payment(
+                    record.order_id.partner_id
+                )
+                record.update({"agent_ids": agent_ids})
+            else:
+                # Regular SOL
+                if not record.commission_free and record.product_id:
```
pedrobaeza: This is the same as in invoices: call super instead.

sale_commission_advance/models/account_move.py
ilyasProgrammer: to make it required=False

pedrobaeza: Then just put such value, not the whole definition. Anyway, it's weird to have a commission line... without commission!

sale_commission_advance/models/account_move.py
ilyasProgrammer: you will get  'bool' object is not iterable

ilyasProgrammer:
@pedrobaeza updated

sale_commission_advance/models/account_move.py
pedrobaeza: OK, I see, then it would be:

```suggestion
        dp_list = self.invoice_line_ids.sale_line_ids.is_downpayment
        return dp_list and any(dp_list)
```

pedrobaeza: This is more readable, but also more efficient.

sale_commission_advance/models/account_move.py
ilyasProgrammer: Amount calculation is not based on commission_id in case of down payment. That's why it's ok to have commission line without commission but with amount. But anyway i've removed this from code. Now commission_id will be there, but amount is calculated as percent of down payment.

pedrobaeza (approved):

PR #387, opened 2022-11-21
[15.0][MIG+REF] sale_commission: Refactorization to split and extract the base

pedrobaeza:
Restarting from the excellent work done by Quartile, adding these extra things:

- Squash of the administrative commits.
- Make specific top-level menu and permissions for commissions.
- Don't depend on account for base module.
- Reorganize elements.
- Change some parts of the code to fit better ORM tools.
- Adjust permissions in children modules.

@Tecnativa TT36983


pedrobaeza:
@yostashiro @AungKoKoLin1997 the last steps I talked about last Friday are done and you can see the result here. Tell me if it's OK for you for finally merging this long awaited refactorization.

/ocabot migration sale_commission

yostashiro:
@pedrobaeza Thanks for the PR.  I was thinking that vendor bill creation part would be separated from `account_commission` to achieve finer granularity, but I think we are fine with the proposed changes.  We will try and review the details in the coming days.  Thanks again.

pedrobaeza:
Yeah, I thought about that also, but being not too much lines of code, and complying with the dependencies, I let it inside the same one.

pedrobaeza:
@francesco-ooops your review is welcome here.

francesco-ooops:
@pedrobaeza on the functional side, is there something new specific to test?

pedrobaeza:
The new functional scope is:

- Separate application for commissions.
- Manual settlements.
- Limited access from the other applications to commissions.
- Settlement wizard by type.

francesco-ooops (approved):
functionally LGTM,

since this is the base module I guess users would appreciate having settlement type selected by default

image

and a server action for settlement invoicing

image

but these are just ux improvements

AungKoKoLin1997 (approved):
Functionally LGTM
In my opinion, I think enterprise users may confuse a little bit for home app screen UX. Two apps related with accounting will be displayed in home screen. But just a minor thing.

Screenshot from 2022-11-22 09-35-20


Screenshot from 2022-11-22 09-35-40

commission/models/commission.py
```diff
@@ -0,0 +1,54 @@
...
+        inverse_name="commission_id",
+    )
+    active = fields.Boolean(default=True)
+    amount_base_type = fields.Selection(
+        selection=[("gross_amount", "Gross Amount"), ("net_amount", "Net Amount")],
+        string="Base",
+        required=True,
+        default="gross_amount",
+    )
+
```
yostashiro, line 39: Thought we might want `settlement_type` field here to control the source-document specific requirement - for example, `invoice_state` should only be required when `settlement_type` is `sale_invoice`.  Currently `invoice_state` is required (in `account_commisson`) even through it is not used in case there are cases in the future where settlements are done from different models than invoice.

account_commission/models/res_partner.py
```diff
@@ -0,0 +1,17 @@
...
+class ResPartner(models.Model):
+    """Add some fields related to commissions"""
+
+    _inherit = "res.partner"
+
+    settlement_ids = fields.One2many(
+        comodel_name="commission.settlement",
+        inverse_name="agent_id",
+        readonly=True,
+    )
```
yostashiro: I think this is an oversight of our work, but should we move this back to `commission`? I don't think this should be in this module.

commission/models/commission_settlement.py
```diff
@@ -0,0 +1,139 @@
...
+        selection=[("manual", "Manual")],
+        default="manual",
+        readonly=True,
+        required=True,
+        help="The source of the settlement, e.g. 'Sales invoice', 'Sales order', "
+        "'Purchase order'...",
+    )
+    can_edit = fields.Boolean(
+        compute="_compute_can_edit",
+        help="Technical field for determining if user can" " edit settlements",
```
yostashiro:
```suggestion
        help="Technical field for determining if user can edit settlements",
```

commission/wizards/commission_make_settle.py
```diff
@@ -0,0 +1,152 @@
...
+    agent_ids = fields.Many2many(
+        comodel_name="res.partner", domain="[('agent', '=', True)]"
+    )
+    settlement_type = fields.Selection(selection=[], required=True)
+    can_settle = fields.Boolean(
+        compute="_compute_can_settle",
+        help="Technical field for improving UX when no extra *commission is installed.",
+    )
+
+    def _compute_can_settle(self):
```
yostashiro, line 26:
```suggestion
    @api.depends("date_to")
    def _compute_can_settle(self):
```
Wasn't totally sure if this suggestion would be valid, but compute doesn't seem to run without a trigger.  Here is an image from runboat.
image

account_commission/views/account_commission_settlement_view.xml
```diff
@@ -0,0 +1,32 @@
...
+        <field name="inherit_id" ref="commission.view_settlement_line_tree" />
+        <field name="arch" type="xml">
+            <xpath expr="field[@name='agent_id']" position="after">
+                <field
+                    name="invoice_line_id"
+                    attrs="{'invisible': [('parent.settlement_type', '=', 'invoice')]}"
+                />
+            </xpath>
+        </field>
+    </record>
```
yostashiro: We should probably move these inside `commission_settlement_views.xml`.

account_commission/views/res_partner.xml
```diff
@@ -0,0 +1,34 @@
...
+                            <field name="date_from" />
+                            <field name="date_to" />
+                            <field name="total" sum="Settled total" />
+                            <field name="state" />
+                        </tree>
+                    </field>
+                </group>
+            </xpath>
+        </field>
+    </record>
```
yostashiro: Would it be more appropriate to move this record to `commission` and add record rules to make `sale_invoice` settlement records only visible to `account.group_account_invoice`, now that `manual` settlement is allowed?

commission/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,38 @@
+For adding commissions:
+
+#. Go to *Invoicing > Configuration > Commission Management > Commission types*.
```
yostashiro:
```suggestion
#. Go to *Commissions > Configuration > Commission types*.
```

commission/readme/USAGE.rst
```diff
@@ -0,0 +1,35 @@
...
+
+#. Go to *Invoicing > Customers > Customers* or *Contacts > Contacts*.
+#. Edit or create a new record.
+#. On "Sales & Purchases" page, you will see a field called "Agents" where
+   they can be added. You can put the number of agents you want, but you can't
+   select specific commission for each partner in this base module.
+
+For settling the commissions to agents:
+
+#. Go to *Invoicing > Vendors > Commissions > Settle commissions*.
```
yostashiro:
```suggestion
#. Go to *Commissions > Settlements > Settle Commissions*.
```

commission/readme/USAGE.rst
```diff
@@ -0,0 +1,35 @@
...
+   * Product. It should be a service product for being coherent.
+   * Journal: To be selected between existing purchase journals.
+   * Date: If you want to choose a specific invoice date. You can leave it
+     blank if you prefer.
+   * Settlements: For selecting specific settlements to invoice. You can leave
+     it blank as well for invoicing all the pending settlements.
+
+#. If you want to invoice a specific settlement, you can navigate to it in
+   *Invoicing > Vendors > Settlements*, and click on "Make invoice"
+   button.
```
yostashiro:
```suggestion
```

account_commission/readme/USAGE.rst
```diff
@@ -0,0 +1,17 @@
...
+   icon with several persons represented, next to the "Commission" field in the
+   list. This icon will be available only if the line hasn't been invoiced yet.
+#. If you have configured your system for editing lines in a popup window,
+   agents will appear also in this window.
+#. The agents icon will be in this ocassion visible when the line hasn't been
+   settled.
+#. You have a button "Regenerate agents" on the bottom of the page
+   "Invoice Lines" for forcing a recompute of all agents from the partner setup.
+   This is needed for example when you have changed the partner on the
+   invoice having already inserted lines.
```
yostashiro, line 17:
```suggestion
   invoice having already inserted lines.

For invoicing the settlements (only for external agents):

#. Go to *Invoicing > Commissions > Invoices Settlements* (or *Commissions > Settlements > Settlements*) and open a settlement record.
#. Click "Make invoice" button, and on the window that appears, you can select following data:

   * Product. It should be a service product for being coherent.
   * Journal: To be selected between existing purchase journals.
   * Date: If you want to choose a specific invoice date. You can leave it
     blank if you prefer.
   * Settlements: For selecting specific settlements to invoice. You can leave
     it blank as well for invoicing all the pending settlements.
```

account_commission/models/account_move.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def button_cancel(self):
+        """Check settled lines and put settlements associated to the invoices in
+        exception.
+        """
+        if any(self.mapped("invoice_line_ids.any_settled")):
+            raise exceptions.ValidationError(
+                _("You can't cancel an invoice with settled lines"),
+            )
+        self.mapped("line_ids.settlement_id").write({"state": "except_invoice"})
```
yostashiro, line 59: I think we need to take care of unlink cases as well.

yostashiro (changes requested):

elvise:
@pedrobaeza  I think "Agents" instead of "Commissions" makes more sense

C2B53692-6466-4A65-BA15-C214FCB00E69

yostashiro:
@pedrobaeza We would like to expedite this migration due to our project situation.  If you are too busy to update this PR soon, we will try following up with another PR, adding a commit on top of yours.  Would you like us to do so?  In that case, can you please comment on each suggested point with your opinion/judgment so that your ideas will be reflected in our update.

pedrobaeza:
Sorry, and thanks for the quality feedback. I'm on it.

commission/models/commission.py
pedrobaeza: I was modifying this, but then realize that this field is for controlling about which price is referred to, and this is something universal.

Anyway, I'm adding the settlement_type for being able to restrict a commission to a certain type of operation, or leave it blank for allowing in all of them.

pedrobaeza: Already pushed in 433a3a2a8a0fdde46395e4e5dfc56f4da232d8af this change for being reviewed while I do the rest of the changes.

pedrobaeza:
> I think "Agents" instead of "Commissions" makes more sense

Not on the main menu, as that one contains all the commission related things, including the agents.

pedrobaeza:
> In my opinion, I think enterprise users may confuse a little bit for home app screen UX. Two apps related with accounting will be displayed in home screen. But just a minor thing.

@AungKoKoLin1997 fixed using this trick: https://github.com/OCA/commission/pull/387/files#diff-0be2abbed382f1580ef0a0f9ee00f07e299ef4d714aeecad3d64b1db2ebefab9R8

account_commission/views/res_partner.xml
pedrobaeza: Security rules are already set, so only moving it is needed.

account_commission/models/account_move.py
pedrobaeza: In fact, for unlinking or cancelling, you need to be in draft state, so I have changed the overrided method to `button_draft`.

pedrobaeza:
Everything sorted out or commented. Please tell me if anything more is found.

account_commission/models/commission.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class Commission(models.Model):
+    _inherit = "commission"
+
+    invoice_state = fields.Selection(
+        [("open", "Invoice Based"), ("paid", "Payment Based")],
+        string="Invoice Status",
+        required=True,
```
yostashiro:
```suggestion
```

account_commission/views/commission_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="commission_form" model="ir.ui.view">
+        <field name="name">commissions form</field>
+        <field name="model">commission</field>
+        <field name="inherit_id" ref="commission.commission_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='commision_params']" position="inside">
+                <field name="invoice_state" />
```
yostashiro:
```suggestion
                <field name="invoice_state" attrs="{'invisible': [('settlement_type', '!=', 'sale_invoice')], 'required': [('settlement_type', '=', 'sale_invoice')]}" />
```

account_commission/wizards/wizard_invoice.xml
```diff
@@ -0,0 +1,46 @@
...
+        </field>
+    </record>
+    <record id="action_commission_make_invoices" model="ir.actions.act_window">
+        <field name="name">Create Commission Invoices</field>
+        <field name="res_model">commission.make.invoice</field>
+        <field name="view_mode">form</field>
+        <field name="target">new</field>
+        <field name="binding_model_id" ref="model_commission_make_invoice" />
+    </record>
+</odoo>
```
yostashiro, line 54:
```suggestion
    <menuitem
        id="menu_commission_make_invoices"
        parent="menu_invoicing_commission"
        action="action_commission_make_invoices"
        sequence="20"
    />
</odoo>
```

yostashiro: Somehow the menu item was missing.

account_commission/readme/USAGE.rst
```diff
@@ -0,0 +1,33 @@
...
+
+For invoicing the settlements (only for external agents):
+
+#. Go to *Invoicing > Vendors > Create commission invoices*.
```
yostashiro:
```suggestion
#. Go to *Invoicing > Commissions > Create commission invoices*.
```

account_commission/views/commission_settlement_views.xml
```diff
@@ -0,0 +1,73 @@
...
+                    groups="account.group_account_invoice"
+                />
+            </field>
+            <xpath
+                expr="//field[@name='line_ids']/tree/field[@name='date']"
+                position="after"
+            >
+                <field
+                    name="invoice_line_id"
+                    attrs="{'column_invisible': [('parent.settlement_type', '!=', 'invoice')]}"
```
yostashiro:
```suggestion
                    attrs="{'column_invisible': [('parent.settlement_type', '!=', 'sale_invoice')]}"
```

account_commission/views/commission_settlement_views.xml
```diff
@@ -0,0 +1,73 @@
...
+    <record model="ir.ui.view" id="view_settlement_line_tree_inherit">
+        <field name="name">Settlement lines</field>
+        <field name="model">commission.settlement.line</field>
+        <field name="inherit_id" ref="commission.view_settlement_line_tree" />
+        <field name="arch" type="xml">
+            <xpath expr="field[@name='agent_id']" position="after">
+                <field
+                    name="invoice_line_id"
+                    attrs="{'invisible': [('parent.settlement_type', '=', 'invoice')]}"
+                />
```
yostashiro:
```suggestion
                <field name="invoice_line_id" />
```

account_commission/models/account_move.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def button_draft(self):
+        """Check settled lines and put settlements associated to the invoices in
+        exception.
+        """
+        if any(self.mapped("invoice_line_ids.any_settled")):
+            raise exceptions.ValidationError(
+                _("You can't cancel an invoice with settled lines"),
+            )
+        self.mapped("line_ids.settlement_id").write({"state": "except_invoice"})
```
yostashiro, line 59:
```suggestion
        self.mapped("line_ids.settlement_id").write({"state": "invoiced"})
```
We should not do this here for button_draft() as it makes the settlement stuck at `except_invoce` state with no way to re-create a vendor bill for the settlement. IMO, no state update should be done on settlement when vendor bill state is changed from 'posted' to 'draft'. Setting `except_invoice` on the settlement should be done in button_cancel() as it did in 14.0. Instead, in button_draft(), we may actually update the settlement to `invoiced` state to handle the cases where vendor bill state is changed from 'cancel' to 'draft'.  Unlink case should still be handled to update the state of the settlement back to `settled`, because, otherwise, the settlement state will be stuck in the `except_invoice` state when 'cancel' invoice is deleted.

yostashiro (commented):
@pedrobaeza Thank you very much for your fast follow-up. I did some functional tests and there are a few more points to take care of.

account_commission/views/commission_views.xml
pedrobaeza: If no `settlement_type` set (available for all type of commissions), it should be also possible to be set. I'm going to put only the required one, but not to hide it.

account_commission/wizards/wizard_invoice.xml
pedrobaeza: This one doesn't work through wizard, but it's an action bound to the settlements. The flow is:

- You select one or several settlements.
- Click on Action > Create Commission Invoices.
- The wizard appears.

pedrobaeza: OK, but I see the other flow you want. I'll implement both.

account_commission/models/account_move.py
pedrobaeza: OK, I have put it back this on cancel method, but added a way to reset a settlement in except state to draft for being re-invoiced.

pedrobaeza:
Changes pushed.

account_commission/models/account_move.py
```diff
@@ -0,0 +1,233 @@
...
+        return super().action_post()
+
+    def button_cancel(self):
+        """Check settled lines and put settlements associated to the invoices in
+        exception.
+        """
+        if any(self.mapped("invoice_line_ids.any_settled")):
+            raise exceptions.ValidationError(
+                _("You can't cancel an invoice with settled lines"),
+            )
```
yostashiro, line 58: I was thinking that this check should stay in button_draft() to ensure the consistency.  I think doing this check in cancel step is a remnant of 12.0 in which cancel was the step to take to revert the posted invoice, which has changed in 13.0.  IMO, it should be done in button_draft() now, but it's OK if it's handled in another PR also.

yostashiro (approved):
Code review. 👍 Thank you very much!

account_commission/models/account_move.py
pedrobaeza: Yes, that's was my initial thought, but in fact being a draft means the same as other areas: a draft invoice in sales means that the order is not invoiceable while not cancelled, so this one is a good balance being able to re-invoice the settlement clicking the button.

pedrobaeza:
Let's go with this! I will add later the migration scripts for transforming 14.0 datamodel to this one:

/ocabot merge major

mgielissen:
No migrations script?

pedrobaeza:
There are 2 options:

- You do them and contribute it.
- Wait until I do it, but not now: https://github.com/OCA/commission/pull/387#issuecomment-1337303688. They are not needed for vanilla 15.0 DBs, so not a requirement for merging.

hildickethan:
@pedrobaeza From what I see in the migration commit, the migration script would be primarily renaming models and fields. I suppose the biggest part is how to handle the module splitting.
~~Should an Odoo with 14.0 `sale_commission` that migrates to 15.0 have `account_commission` auto-installed too? Since it's part of the module they had, if they don't use that part then it can be uninstalled later.~~ I didn't see that it depends on it, I suppose that's not a problem then

I'm looking to migrate this to 16.0, but I don't really have a good test subject to help with this migration script right now, do you have anything planned for it? 

Edit: Now that the PR is done and I mentioned you again, probably better to reply there https://github.com/OCA/commission/pull/394

pedrobaeza:
Yes, basically is to rename some tables, models, columns and fields, and set some values. I will do this on February.

PR #390, opened 2022-12-06
[FIX][15.0] account_commission : Not working "Reset to draft" button

kanda999:
The state in the commission.settlement model is settled, not draft.

I get an error when I press the "Reset to draft" button.


yostashiro (approved):
I'd personally prefer not to change the method name, but I guess it's OK.

pedrobaeza (changes requested):
Yes, please only change the written value, not all the method + string names

pedrobaeza (approved):

PR #391, opened 2022-12-07
[15.0][IMP] account_commission: link settlements and vendor bill that is created from settlements

AungKoKoLin1997:
After installing this module, there will be a settlement link button in vendor bill that is created for settlement in the form view and tree view like below image.
Screenshot from 2022-12-07 17-27-12

Screenshot from 2022-12-07 17-27-23

### DIFF
--- a/account_commission/models/account_move.py
+++ b/account_commission/models/account_move.py
@@ -21,6 +21,31 @@ class AccountMove(models.Model):
         compute="_compute_agents",
         search="_search_agents",
     )
+    settlement_count = fields.Integer(compute="_compute_settlement")
+    settlement_ids = fields.One2many(
+        "commission.settlement",
+        string="Settlements",
+        compute="_compute_settlement",
+    )
+
+    def action_view_settlement(self):
+        xmlid = "commission.action_commission_settlement"
+        action = self.env["ir.actions.actions"]._for_xml_id(xmlid)
+        action["context"] = {}
+        settlements = self.mapped("settlement_ids")
+        if not settlements or len(settlements) > 1:
+            action["domain"] = [("id", "in", settlements.ids)]
+        elif len(settlements) == 1:
+            res = self.env.ref("commission.view_settlement_form", False)
+            action["views"] = [(res and res.id or False, "form")]
+            action["res_id"] = settlements.id
+        return action
+
+    def _compute_settlement(self):
+        for invoice in self:
+            settlements = invoice.invoice_line_ids.settlement_id
+            invoice.settlement_ids = settlements
+            invoice.settlement_count = len(settlements)
 
     @api.depends("partner_agent_ids", "invoice_line_ids.agent_ids.agent_id")
     def _compute_agents(self):
--- a/account_commission/models/commission_settlement.py
+++ b/account_commission/models/commission_settlement.py
@@ -30,6 +30,7 @@ class CommissionSettlement(models.Model):
         readonly=True,
     )
     invoice_id = fields.Many2one(
+        string="Generated Invoice",
         store=True,
         comodel_name="account.move",
         compute="_compute_invoice_id",
@@ -44,7 +45,9 @@ class CommissionSettlement(models.Model):
     @api.depends("invoice_line_ids")
     def _compute_invoice_id(self):
         for record in self:
-            record.invoice_id = record.invoice_line_ids[:1].move_id
+            record.invoice_id = record.invoice_line_ids.filtered(
+                lambda x: x.parent_state != "cancel"
+            )[:1].move_id
 
     def action_cancel(self):
         """Check if any settlement has been invoiced."""
--- a/account_commission/views/account_move_views.xml
+++ b/account_commission/views/account_move_views.xml
@@ -41,6 +41,21 @@
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
         <field name="arch" type="xml">
+            <xpath expr="//button[@name='open_created_caba_entries']" position="after">
+                <button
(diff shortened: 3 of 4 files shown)


yostashiro (commented):
@AungKoKoLin1997 Can you please add tests and also share images of the adjusted views.

account_commission/models/account_move.py
```diff
@@ -21,6 +21,41 @@ class AccountMove(models.Model):
...
+            form_view = [(self.env.ref("commission.view_settlement_form").id, "form")]
+            if "views" in action:
+                action["views"] = form_view + [
+                    (state, view) for state, view in action["views"] if view != "form"
+                ]
+            else:
+                action["views"] = form_view
+            action["res_id"] = settlements.id
+        else:
+            action = {"type": "ir.actions.act_window_close"}
```
pedrobaeza: No need of this. Simply don't populate any action in this case, but avoid this `else` that is not covered in tests.

account_commission/models/account_move.py
```diff
@@ -21,6 +21,41 @@ class AccountMove(models.Model):
...
+
```
pedrobaeza: Don't add empty lines in the middle of a method.

pedrobaeza (commented):
What about the inverse one: settlement > invoice? Maybe a note logged on the settlement saying that the invoice X has been generated with the possibility to navigate from it?

AungKoKoLin1997:
> What about the inverse one: settlement > invoice? Maybe a note logged on the settlement saying that the invoice X has been generated with the possibility to navigate from it?

@pedrobaeza For the inverse one, I think we have invoice field and there is no problem for lacking button in form.

Screen Shot 2022-12-07 at 21 17 26

pedrobaeza:
OK, agreed. What you can change is the label to "Generated invoice".

AungKoKoLin1997:
> OK, agreed. What you can change is the label to "Generated invoice".

Ok.  I will change label

pedrobaeza (approved):

pedrobaeza:
Please change the commit message to `[IMP] account_commission: Add link invoice > settlement`

pedrobaeza (approved):

account_commission/models/account_move.py
```diff
@@ -21,6 +21,33 @@ class AccountMove(models.Model):
         compute="_compute_agents",
         search="_search_agents",
     )
+    settlement_count = fields.Integer(compute="_compute_settlement")
+    settlement_ids = fields.Many2many(
```
yostashiro: Should this field technically be one2many given that there should be only one valid vendor bill for a settlement?

We may take this chance to update _compute_invoice_id() of commission.settlement to exclude cancelled vendor bills as well.
https://github.com/oca/commission/blob/b6715536aa7c115b9283d07e10f7ffae51268ee2/account_commission/models/commission_settlement.py#L44-L47

@pedrobaeza Any thoughts on this?

pedrobaeza: OK for me.

account_commission/views/account_move_views.xml
```diff
@@ -111,4 +126,22 @@
             </field>
         </field>
     </record>
+    <record id="view_in_invoice_bill_tree_inherit" model="ir.ui.view">
```
yostashiro, line 119: Please move this record after `view_move_form` record.

account_commission/models/commission_settlement.py
```diff
@@ -30,6 +30,7 @@ class CommissionSettlement(models.Model):
         readonly=True,
     )
     invoice_id = fields.Many2one(
+        string="Generated invoice",
```
yostashiro:
```suggestion
        string="Generated Invoice",
```

account_commission/models/commission_settlement.py
```diff
@@ -30,6 +30,7 @@ class CommissionSettlement(models.Model):
...
         store=True,
         comodel_name="account.move",
         compute="_compute_invoice_id",
```
yostashiro, line 36: Please update the content of the compute method to:
```python
        for record in self:
            record.invoice_id = record.invoice_line_ids.filtered(lambda x: x.parent_state != "cancel")[:1].move_id
```

account_commission/models/account_move.py
```diff
@@ -21,6 +21,33 @@ class AccountMove(models.Model):
...
+        "commission.settlement",
+        string="Settlements",
+        compute="_compute_settlement",
+        copy=False,
+    )
```
yostashiro:
```suggestion
    settlement_ids = fields.One2many(
        "commission.settlement",
        string="Settlements",
        compute="_compute_settlement",
    )
```

account_commission/models/account_move.py
```diff
@@ -21,6 +21,33 @@ class AccountMove(models.Model):
...
+        settlements = self.mapped("settlement_ids")
+        if not settlements or len(settlements) > 1:
+            action["domain"] = [("id", "in", settlements.ids)]
+        elif len(settlements) == 1:
+            res = self.env.ref("commission.view_settlement_form", False)
+            action["views"] = [(res and res.id or False, "form")]
+            action["res_id"] = settlements.id
+        return action
+
+    @api.depends("invoice_line_ids")
```
yostashiro:
```suggestion
```

yostashiro (changes requested):
Mostly nitpicking.

yostashiro (approved):
Code review. 👍

pedrobaeza:
Please squash commits into one.

AungKoKoLin1997:
> Please squash commits into one.

Done!

PR #394, opened 2022-12-26
[16.0][MIG] commission

hildickethan:
- Polished some code for better time complexities and using more modern features like removing `mapped` for relational fields

@pedrobaeza No migration script for the 14.0 -> 15.0 yet either, but I will probably look into doing it if no one else has within a couple months whenever we get around to migrating the production DB


pedrobaeza:
Please do only one module per PR, starting with `commission`.

hildickethan:
@pedrobaeza Split into 3 PRs, https://github.com/OCA/commission/pull/395 https://github.com/OCA/commission/pull/396

commission/models/commission_mixin.py
```diff
@@ -37,13 +37,15 @@ def _prepare_agent_vals(self, agent):
...
-            agents = agents.filtered(
-                lambda x: not x.commission_id.settlement_type
-                or x.commission_id.settlement_type == settlement_type
+        return [
+            (0, 0, self._prepare_agent_vals(agent))
+            for agent in partner.agent_ids
+            if settlement_type
+            and (
+                not agent.commission_id.settlement_type
+                or agent.commission_id.settlement_type == settlement_type
```
pedrobaeza: This is not expressing the same condition as before, and I think this is less readable.

commission/models/commission_settlement.py
```diff
@@ -69,7 +69,7 @@ def _default_company_id(self):
     @api.depends("line_ids", "line_ids.settled_amount")
     def _compute_total(self):
         for record in self:
-            record.total = sum(record.mapped("line_ids.settled_amount"))
+            record.total = sum(record.line_ids.mapped("settled_amount"))
```
pedrobaeza: No need of this diff. It works the same.

commission/reports/report_settlement_templates.xml
```diff
@@ -6,15 +6,15 @@
                 <t t-call="web.internal_layout">
                     <div class="page">
                         <div class="row mt32 mb32">
-                            <div class="col-xs-3" t-if="o.agent_id">
+                            <div class="col-3" t-if="o.agent_id">
```
pedrobaeza, line 9: Why changing this?

commission/wizards/commission_make_settle.py
```diff
@@ -104,19 +104,20 @@ def action_settle(self):
...
             # Get non settled elements
             agent_lines = self._get_agent_lines(agent, date_to_agent)
-            for company in agent_lines.mapped("company_id"):
-                agent_lines_company = agent_lines.filtered(
-                    lambda r: r.object_id.company_id == company
-                )
+            agent_lines_by_company = {}
+            for line in agent_lines:
+                agent_lines_by_company.setdefault(line.company_id, line)
+                agent_lines_by_company[line.company_id] |= line
```
pedrobaeza: This set operation is slower than the filtered, so please keep the other.

commission/reports/report_settlement_templates.xml
hildickethan: `col-xs-*` was replaced with `col-*` in bootstrap 4

pedrobaeza: Ok

commission/wizards/commission_make_settle.py
hildickethan: Do you know why this is slower? Because it's true that I didn't actually test the time differences until now and I can't really make the new solution better even in bigger `agent_lines` lengths (~1000) like I expected, with prefetching doing the bulk of the work already.

I'm open to change it back but I will explain that the idea behind this is that iterating once to group lines by company and once to iterate on the grouped lines being O(2n) sounds much better to me than the `filtered` that iterates through every line inside the `for`, making it O(n^n).
I saw a very similar case recently relating to a website method that drastically reduced the execution time from 60 to 20 seconds when having ~800 lines with a change nearly identical to this, which is why it seemed sensible to do

commission/models/commission_mixin.py
hildickethan: The idea behind this change was also related to time complexity like my other comment, but I agree that it is harder to read and since this method seems to only be used for single lines, I will change it back

commission/models/commission_settlement.py
hildickethan: I think it looks cleaner like this :thinking: Should I not change these then?

pedrobaeza: The principle is to reduce the diff at maximum, specially with cosmetic changes.

commission/wizards/commission_make_settle.py
pedrobaeza: The cache makes successive `filtered` operations to run smooth.

The problem with set operations is that the decision behind recordset operations is to copy the recordset structure, and in the copy, append the new record. This operation can be memory and time consuming over the other one. You will notice it with a bigger set.

hildickethan: I see, I reverted it back to the original

hildickethan:
Reverted the unnecessary changes, after this PR is merged/approved I will adjust the other 2 PRs based on the comments for this one

pedrobaeza (approved):
Congratulations on the diligence!

commission/views/commission_settlement_views.xml
```diff
@@ -0,0 +1,178 @@
...
+                            colspan="2"
+                        >
+                            <tree editable="top">
+                                <field name="date" />
+                                <field name="commission_id" />
+                                <field name="settled_amount" />
+                            </tree>
+                        </field>
+                    </group>
+                    <group class="oe_invoice_lines_tab">
```
yostashiro:
```suggestion
                    <group>
```
I think we should remove this style as it comes from `account` module and doesn't seem to do anything here.

yostashiro (commented):
Thanks for the migration. Just a minor point.

yostashiro:
Another thing about the settlement form (but this is not a migration issue) - the presentation of `date_from` and `date_to` fields should be fixed so that they are displayed side by side when multi-company is not enabled.  If you can add a separate commit on this, we will backport it to 15.0.

image

commission/views/commission_settlement_views.xml
hildickethan: This was cause the footer showed up squished for me until I put this style on, but removing it doesn't change anything so maybe it was probably some cache that got reset when i applied the style or something, I will remove it

hildickethan: Updated

hildickethan:
> Another thing about the settlement form (but this is not a migration issue) - the presentation of `date_from` and `date_to` fields should be fixed so that they are displayed side by side when multi-company is not enabled. If you can add a separate commit on this, we will backport it to 15.0.
> 

Changed in [[IMP] commission: Align dates in settlement view](https://github.com/OCA/commission/pull/394/commits/6f03761aeb9c212c819e9259a0599227e7bed208). I moved the company below the date for it

Captura desde 2022-12-28 09-07-13
Captura desde 2022-12-28 09-07-27

yostashiro (approved):
Thanks for the updates.

yostashiro:
> Changed in [[IMP] commission: Align dates in settlement view](https://github.com/OCA/commission/pull/394/commits/6f03761aeb9c212c819e9259a0599227e7bed208). I moved the company below the date for it

I was thinking maybe the following style would be appropriate for usability, etc. but I'm fine with your update. Thank you.
https://github.com/oca/ocb/blob/5857bb9f00ebb30cddab256ffd634c3ba54252e4/addons/hr_holidays/views/hr_leave_views.xml#L275-L285

pedrobaeza:
@yostashiro you can add in a successive PR such improvement, and then backport the final result to v15. I'm merging now this:

/ocabot merge nobump

mgielissen:
> * Standard migration to 16.0 with `odoo-module-migrator` tool
> 
>     * Polished some code for better time complexities and using more modern features like removing `mapped` for relational fields
> 
> 
> @pedrobaeza No migration script for the 14.0 -> 15.0 yet either, but I will probably look into doing it if no one else has within a couple months whenever we get around to migrating the production DB

Any news about migration script?

pedrobaeza:
Not needed yet by my part.

hildickethan:
For us we opted to just closing commissions and invoicing them in the previous version and then starting fresh in a new month when migrating and migrating old data later if need be

PR #395, opened 2022-12-27
[16.0][MIG] account_commission

hildickethan:
- Polished some code for better time complexities and using more modern features like removing `mapped` for relational fields
- Didn't do this TODO `# todo or compute agent currency_id?` as I'm not sure what it means, but I'm open to finishing it if I could get some pointers
- Depends on https://github.com/OCA/commission/pull/394 
- Changed Forms in tests to standard creates due to this commits description: https://github.com/odoo/odoo/commit/d8d47f9ff8554f4b39487fd2f13c153c7d6f958d
I found it necessary for things to work properly and pass the existing tests
`* no need to use the Form emulator to build invoices (i.e. EDI, OCR, intercompany, ...)`


hildickethan:
Updated migration commit with feedback from https://github.com/OCA/commission/pull/394 + rebased

account_commission/models/account_move.py
```diff
@@ -143,16 +143,23 @@ def _compute_agent_ids(self):
                 [("invoice_line_id", "=", res.id)]
             )
             for line in settlement_lines:
-                line.date = False
-                line.agent_id = False
-                line.settled_amount = 0.00
-                line.currency_id = False
-                line.commission_id = False
+                line.update(
```
pedrobaeza: Since v13, the efficiency of assigning with active record pattern (dot notation) once each time is the same as using `update` for assigning all in one operation, so no need of this change.

pedrobaeza (approved):

account_commission/models/account_move.py
hildickethan: :thinking: interesting, didn't know that
Reverted

pedrobaeza: Yes, it's due to 2 changes in the ORM layer:

- Computed fields calculation postponed till they are used.
- Deferred writing, done only at the end.

account_commission/models/commission_settlement.py
```diff
@@ -69,52 +68,59 @@ def action_invoice(self):
...
+                        "name": (
+                            "\n"
+                            + _(
+                                "Period: from %(date_from)s to %(date_to)s",
+                                date_from=date_from.strftime(lang.date_format),
+                                date_to=date_to.strftime(lang.date_format),
+                            ),
+                        ),
+                        # todo or compute agent currency_id?
+                        "currency_id": settlement.currency_id.id,
```
yostashiro, line 119: I suppose we can remove these lines?  It's already a related field.

yostashiro (approved):
Code review. 👍

account_commission/models/commission_settlement.py
hildickethan: wasn't sure what to do with these so I left it as is, but removing them should work

pedrobaeza: I don't know in v16, but in previous versions, related fields are filled at the end of the pipeline, so if there's any reference to this on computations, it will be considered as empty, so it's not always a bad idea to preset it. This happens mostly on defaults based on related fields, but not sure if it can happen in other parts.

PR #396, opened 2022-12-27
[16.0][MIG] sale_commission

hildickethan:
- Depends on
  - https://github.com/OCA/commission/pull/394 
  - https://github.com/OCA/commission/pull/395


hildickethan:
Updated migration commit with feedback from https://github.com/OCA/commission/pull/394, pending rebase with merge of https://github.com/OCA/commission/pull/395 for green tests

hildickethan:
Rebased :green_apple:

pedrobaeza (approved):

yostashiro (approved):
LGTM. 👍 

One minor(?) thing - the module now depends on `sale` instead of `sale_management`, and therefore runboat instance does not show the sales menu unless you go to apps and install `sale_management`.  Should we switch the dependency to `sale_management`?

In fact the `sale_commission` module in 14.0 depends on `sale_management`, and in our migration to 15.0, we changed it to `sale` as we thought it was technically not necessary to depend on `sale_management`, however I think it may have been an unnecessary move.

@pedrobaeza Can you please advise?

hildickethan:
> In fact the `sale_commission` module in 14.0 depends on `sale_management`, and in our migration to 15.0, we changed it to `sale` as we thought it was technically not necessary to depend on `sale_management`, however I think it may have been an unnecessary move.

I think this was primarily due to moving the commissions menu out of sales and into its own. To my knowledge the norm is just to depend on `sale` unless you specifically need the dependency for some inheritance

pedrobaeza:
Yes, let's keep it with the minimum technical dependencies:

/ocabot merge nobump

PR #403, opened 2023-02-24
Added translation to spanish for commission and account_commission

JavierVescio:


pedrobaeza:
Please use Weblate: https://odoo-community.org/resources/translate

PR #404, opened 2023-02-26
[14.0][FIX] sale_commission: partner is set previously

etobella:
This wil give you an error when combined with sale_commission_delegated_partner


pedrobaeza (approved):
It takes a bit to understand the problem, but I got it, and can be fast-tracked. Please fw-port it:

/ocabot merge patch

PR #407, opened 2023-03-22
[14.0][IMP] sale_commission - agent lines domain in a specific function

odooNextev:
We need to edit agent lines domain for a specific payment term defined by a module in `l10n-italy` repo and we thought to add a method to return the domain instead including this in the whole function.
In fact, at the moment, we'd overwrite the whole action_settle method breaking the chain of inheritance.

### DIFF
--- a/sale_commission/wizard/wizard_settle.py
+++ b/sale_commission/wizard/wizard_settle.py
@@ -75,7 +75,6 @@ class SaleCommissionMakeSettle(models.TransientModel):
 
     def action_settle(self):
         self.ensure_one()
-        agent_line_obj = self.env["account.invoice.line.agent"]
         settlement_obj = self.env["sale.commission.settlement"]
         settlement_line_obj = self.env["sale.commission.settlement.line"]
         settlement_ids = []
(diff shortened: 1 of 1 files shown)


ilyasProgrammer (approved):
LGTM

francesco-ooops:
@tafaRU good for you?

sale_commission/wizard/wizard_settle.py
```diff
@@ -88,13 +87,8 @@ def action_settle(self):
...
             # Get non settled invoices
-            agent_lines = agent_line_obj.search(
-                [
-                    ("invoice_date", "<", date_to_agent),
-                    ("agent_id", "=", agent.id),
-                    ("settled", "=", False),
-                ],
-                order="invoice_date",
+            agent_lines = self._get_agent_lines(agent, date_to_agent).sorted(
+                key=lambda al: al.invoice_date
```
pedrobaeza: Please don't separate the sort from the search. It's not efficient.

pedrobaeza (commented):
Instead of this, may be better to have the hook on the domain returned, or both.

stenext:
> Instead of this, may be better to have the hook on the domain returned, or both.

You're right @pedrobaeza, but we cannot filter agent lines in our module https://github.com/OCA/l10n-italy/pull/3222 (PR is for 12.0 but we're almost done with 14.0 too) with domain only.

This filter is impossible to do with domain only: https://github.com/OCA/l10n-italy/pull/3222/files#diff-bf51edd38dc84ce836a63ebb9a93c054688047a003d2c06a4e30ec0dca88708eR30

pedrobaeza:
OK, please do the rest of my comments.

stenext:
> OK, please do the rest of my comments.

Ok, Pedro, I did it https://github.com/OCA/commission/pull/407/commits/93940885f5a59795338dafbe634e04bb4d10705d

Let me know if I gotta do something else

pedrobaeza:
OK, please reduce to one commit, and I don't think this can be considered a contribution to be noted on the README, as it's only a commodity for you, but it's not improving the module at all. The tag for this is `[REF]` instead of `[IMP]`.

stenext:
> OK, please reduce to one commit, and I don't think this can be considered a contribution to be noted on the README, as it's only a commodity for you, but it's not improving the module at all. The tag for this is `[REF]` instead of `[IMP]`.

I guess I did everything you told me, let me know

pedrobaeza (approved):

PR #409, opened 2023-03-29
[MIG] [16.0] commission_formula

Nikul-OSI:


commission_formula/__manifest__.py
```diff
@@ -10,7 +10,7 @@
     "summary": "Commissions computed by formulas",
     "author": "Abstract,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/commission",
-    "depends": ["commission"],
+    "depends": ["commission", "account"],
```
pedrobaeza, line 13: This added dependency doesn't seem correct.

pedrobaeza (changes requested):
The module version bumping is not in the correct commit.

commission_formula/__manifest__.py
Nikul-OSI: Here we used move line https://github.com/OCA/commission/pull/409/commits/f141ae9381f3240cc33a477869006759870d5627#diff-628b4901eac43e5fc9b48afa3b145f4d37d09a8ce33340b551f2297524fbc072R11

pedrobaeza: The link is not correctly shown, but I have checked full code and I don't see such dependency.

PR #411, opened 2023-04-11
[16.0][FIX] account_commission: Regenerate agents button invisibility

jguenat:
- xml states and attrs invisible are combined with AND operator which was not correct in this case, we need OR.

### DIFF
--- a/account_commission/views/account_move_views.xml
+++ b/account_commission/views/account_move_views.xml
@@ -89,9 +89,10 @@
                     name="recompute_lines_agents"
                     type="object"
                     string="Regenerate agents"
-                    states="draft"
                     attrs="{'invisible': [
-                        ('move_type', 'not in', ['out_invoice', 'out_refund'])
+                        '|',
+                        ('move_type', 'not in', ['out_invoice', 'out_refund']),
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Please remember to include the version in the PR title (only there, not in the commit message).

pedrobaeza (changes requested):
Please don't mix things in the same PR.

jguenat:
Hello @pedrobaeza 
Is it possible to do a single PR with multiple small fixes in this case?
i can rename the title and it would reduce PR noise?

pedrobaeza:
They are a bit conflicting, as I can agree with one patch, but not with others. I don't agree with the latest ones, and that's why I have raised this concern.

jguenat:
Okay then I'll create separate PRs, also bonus question, is there anything I can do to make codecov/project happy?
How is it possible to have a patch coverage of 100% and projet diminishing?
thanks a lot for your time.

pedrobaeza:
I'm thinking since a bit that Codecov is not being fed with correct data for its analysis, so don't take it very seriously.

pedrobaeza (approved):

PR #364, opened 2022-08-02
[ADD] sale_commission_product_based

SirTakobi:

PR #414, opened 2023-04-17
[14.0][ADD] sale_commission_product_criteria

ilyasProgrammer:
This module allows to set in the same Commission Type different commission rates according to the product on SO/invoice line.

This is made possible since this module adds a new "Rules" type to Commission Type and applies commission rates with the same logic of sale pricelist items.

For example, such a Commission Type can grant:

10% on a specific Product A,
10$ on Product B,
4% on products in Category 1 and
5$ on all other products.

In SO/invoice, system will apply different commissions based on variant/product/category or global, applied hierarchically. This means that for the example above, if product A is assigned to Category 1, commission assigned is 10%, as per variant/product/category/global rule application order.

Furthermore, these commission type items can be accessed and created by a specific menu, to facilitate their management in environments with lots of records.

The form for commission type item can be extended by future modules with further conditions to decide when to apply a specific item.

### DIFF
--- a/sale_commission_product_criteria/__manifest__.py
+++ b/sale_commission_product_criteria/__manifest__.py
@@ -1,0 +1,21 @@
+# © 2023 ooops404
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+{
+    "name": "Sale Commission Product Criteria",
+    "summary": "Advanced commissions rules",
+    "version": "14.0.1.0.1",
+    "author": "Ilyas, Ooops404, Odoo Community Association (OCA)",
+    "maintainers": ["ilyasProgrammer"],
+    "website": "https://github.com/OCA/commission",
+    "category": "Sales Management",
+    "license": "AGPL-3",
+    "depends": ["sale_commission", "web_domain_field"],
+    "data": [
+        "views/views.xml",
+        "security/ir.model.access.csv",
+    ],
+    "demo": ["demo/sale_agent_demo.xml"],
+    "application": False,
+    "installable": True,
+    "auto_install": False,
+}
--- a/sale_commission_product_criteria/models/commission.py
+++ b/sale_commission_product_criteria/models/commission.py
@@ -1,0 +1,267 @@
+# © 2023 ooops404
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.tools import float_repr
+
+
+class SaleCommission(models.Model):
+    _inherit = "sale.commission"
+
+    commission_type = fields.Selection(
+        selection_add=[("product", "Product criteria")],
+        ondelete={"product": "set default"},
+    )
+    item_ids = fields.One2many("commission.item", "commission_id", copy=True)
+
+    def action_unarchive(self):
+        res = super().action_unarchive()
+        items = (
+            self.env["commission.item"]
+            .with_context(active_test=False)
+            .search([("commission_id", "=", self.id)])
+        )
+        if items:
+            items.write({"active": True})
+        return res
+
+    @api.onchange("commission_type")
+    def onchange_commission_type(self):
+        # Prevent commission_type change in certain cases
+        self.check_type_change_allowed_sale()
+        self.check_type_change_allowed_moves()
+
+    def check_type_change_allowed_sale(self):
+        sola_ids = self.env["sale.order.line.agent"].search(
+            [("commission_id", "=", self._origin.id)]
+        )
+        done_so_ids = sola_ids.filtered(lambda x: x.object_id.state in ["done", "sale"])
+        if done_so_ids:
+            raise ValidationError(
+                _(
+                    "There is done Sale Orders with this commission. "
+                    "Commission type change is not allowed."
+                )
+            )
+
+    def check_type_change_allowed_moves(self):
+        aila_ids = self.env["account.invoice.line.agent"].search(
+            [("commission_id", "=", self._origin.id)]
+        )
+        done_move_ids = aila_ids.filtered(
+            lambda x: x.object_id.parent_state == "posted"
+        )
+        if done_move_ids:
+            raise ValidationError(
+                _(
+                    "There is posted Account Move Lines with this commission. "
+                    "Commission type change is not allowed."
+                )
+            )
+
+
+class CommissionItem(models.Model):
+    _name = "commission.item"
+    _description = "Commission Item"
+    _order = "applied_on, based_on, categ_id desc, id desc"
+
+    sequence = fields.Integer(default=10)
+    active = fields.Boolean(default=True)
+    commission_id = fields.Many2one(
+        "sale.commission",
+        string="Commission Type",
+        domain=[("commission_type", "=", "product")],
+        required=True,
+    )
+    use_pricelist = fields.Boolean()
+    pricelist_id = fields.Many2one("product.pricelist")
+    product_tmpl_id = fields.Many2one(
+        "product.template",
+        "Product",
+        ondelete="cascade",
+        check_company=True,
+        help="Specify a template if this rule only applies to one "
+        "product template. Keep empty otherwise.",
+    )
+    product_id = fields.Many2one(
+        "product.product",
+        "Product Variant",
+        ondelete="cascade",
+        check_company=True,
+        help="Specify a product if this rule only applies "
+        "to one product. Keep empty otherwise.",
+    )
+    categ_id = fields.Many2one(
+        "product.category",
+        "Product Category",
+        ondelete="cascade",
+        help="Specify a product category if this rule only applies to "
+        "products belonging to this category or its children categories. "
+        "Keep empty otherwise.",
+    )
+    based_on = fields.Selection(
+        [("sol", "Any Sale Order Line")],
+        string="Based On",
+        required=True,
+        default="sol",
+    )
+    applied_on = fields.Selection(
+        [
+            ("3_global", "All Products"),
+            ("2_product_category", "Product Category"),
+            ("1_product", "Product"),
+            ("0_product_variant", "Product Variant"),
+        ],
+        "Apply On",
+        default="3_global",
+        required=True,
+        help="Commission Item applicable on selected option",
+    )
+    commission_type = fields.Selection(
+        [("fixed", "Fixed"), ("percentage", "Percentage")],
+        index=True,
+        default="fixed",
+        required=True,
+    )
+    fixed_amount = fields.Float("Fixed Amount", digits="Product Price")
+    percent_amount = fields.Float("Percentage Amount")
+    company_id = fields.Many2one(
+        "res.company",
+        "Company",
+        default=lambda self: self.env.company,
+        readonly=True,
+    )
+    currency_id = fields.Many2one(
+        "res.currency",
+        related="company_id.currency_id",
+        readonly=True,
+    )
+    name = fields.Char(
+        "Name",
+        compute="_compute_commission_item_name_value",
+        help="Explicit rule name for this commission line.",
+    )
+    commission_value = fields.Char(
+        "Value",
+        compute="_compute_commission_item_name_value",
+    )
+
+    @api.depends(
+        "applied_on",
+        "categ_id",
+        "product_tmpl_id",
+        "product_id",
+        "commission_type",
+        "fixed_amount",
+        "percent_amount",
+    )
+    def _compute_commission_item_name_value(self):
+        for item in self:
+            if item.categ_id and item.applied_on == "2_product_category":
+                item.name = _("Category: %s") % (item.categ_id.display_name)
+            elif item.product_tmpl_id and item.applied_on == "1_product":
+                item.name = _("Product: %s") % (item.product_tmpl_id.display_name)
+            elif item.product_id and item.applied_on == "0_product_variant":
+                item.name = _("Variant: %s") % (
+                    item.product_id.with_context(
+                        display_default_code=False
+                    ).display_name
+                )
+            else:
+                item.name = _("All Products")
+
+            if item.commission_type == "fixed":
+                decimal_places = self.env["decimal.precision"].precision_get(
+                    "Product Price"
+                )
+                if item.currency_id.position == "after":
+                    item.commission_value = "%s %s" % (
+                        float_repr(
+                            item.fixed_amount,
+                            decimal_places,
+                        ),
+                        item.currency_id.symbol or "",
+                    )
+                else:
+                    item.commission_value = "%s %s" % (
+                        item.currency_id.symbol or "",
+                        float_repr(
+                            item.fixed_amount,
+                            decimal_places,
+                        ),
+                    )
+            elif item.commission_type == "percentage":
+                item.commission_value = str(item.percent_amount) + " %"
+
+    @api.constrains("product_id", "product_tmpl_id", "categ_id")
+    def _check_product_consistency(self):
+        for item in self:
+            if item.applied_on == "2_product_category" and not item.categ_id:
+                raise ValidationError(
+                    _(
+                        "Please specify the category for which this rule should "
+                        "be applied"
+                    )
+                )
+            elif item.applied_on == "1_product" and not item.product_tmpl_id:
+                raise ValidationError(
+                    _(
+                        "Please specify the product for which this rule should "
+                        "be applied"
+                    )
+                )
+            elif item.applied_on == "0_product_variant" and not item.product_id:
+                raise ValidationError(
+                    _(
+                        "Please specify the product variant for "
+                        "which this rule should be applied"
+                    )
+                )
+
+    @api.onchange("product_id")
+    def _onchange_product_id(self):
+        has_product_id = self.filtered("product_id")
+        for item in has_product_id:
+            item.product_tmpl_id = item.product_id.product_tmpl_id
+        if self.env.context.get("default_applied_on", False) == "1_product":
+            # If a product variant is specified, apply on variants instead
+            # Reset if product variant is removed
+            has_product_id.update({"applied_on": "0_product_variant"})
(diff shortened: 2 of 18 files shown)


francesco-ooops (approved):
Functional review ok!

GSLabIt (approved):

francesco-ooops:
@pedrobaeza looking forward to your opinion on this one :)

pedrobaeza:
The name is very weird for the description given in the main comment. The same we have `sale_commission_formula`, you can have `sale_commission_pricelist_criteria` if I understand correctly the purpose.

There's no information on the README about how to configure the rules, so I can't say more.

sale_commission_type_lines/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module adds a new "type" in Commission Type, named "Rules", allowing to set conditions on commission based on the product according to the same logic of sale pricelist items.
+
+System will apply different commissions based on variant/product/category or global, applied hierarchically.
+
+These commission type items can be accessed and created by a specific menu, to facilitate their management in environments with lots of records.
+
+The form for commission type item can be extended by future modules with further conditions to decide when to apply a specific item.
```
francesco-ooops: This module allows to set in the same Commission Type different commission rates according to the product on SO/invoice line.

This is made possible since this module adds a new "Rules" type to Commission Type and applies commission rates with the same logic of sale pricelist items.

For example, such a Commission Type can grant:

10% on a specific Product A,
10$ on Product B,
4% on products in Category 1 and
5$ on all other products.

In SO/invoice, system will apply different commissions based on variant/product/category or global, applied hierarchically. This means that for the example above, if product A is assigned to Category 1, commission assigned is 10%, as per variant/product/category/global rule application order.

Furthermore, these commission type items can be accessed and created by a specific menu, to facilitate their management in environments with lots of records.

The form for commission type item can be extended by future modules with further conditions to decide when to apply a specific item.

sale_commission_type_lines/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
+To use features of this module, you need to:
+
+#. Go to Sales > Commission Management > Commission Types.
+#. Create or update existing record.
+#. Create multiple rules. Select *Apply On* variant. Fill up other fields.
+#. Rest flow is according to OCA sale_commission module.
```
francesco-ooops: To use features of this module, you need to:

#. Go to Sales > Commission Management > Commission Types.
#. Create a Commission Type with type = "Rules".
#. Create multiple rules based on variant/product/category or global
#. These rules will be sorted according to the same logic of sale pricelist.
#. Rest flow is according to OCA sale_commission module.

francesco-ooops (commented):
@pedrobaeza does this proposed change in description/usage clarify what this module is about?

I don't think sale_commission_type_lines is such an unrelated name as we're actually adding lines to commission types, but we're open to any suggestion :)

Thanks!

pedrobaeza:
Now I understand better the purpose of the module, and the best name for it IMO is `sale_commission_product_criteria`.

francesco-ooops:
@pedrobaeza should be good to review now!

francesco-ooops:
@SirTakobi I think you were trying to achieve the same result with #364 , if you want to take a look :)

sale_commission_product_criteria/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# © 2023 ooops404
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+{
+    "name": "Sale Commission Type Lines",
+    "summary": "Advanced commissions rules",
+    "version": "14.0.1.0.1",
+    "author": "Ilyas," "Ooops404," "Odoo Community Association (OCA)",
```
pedrobaeza:
```suggestion
    "author": "Ilyas, Ooops404, Odoo Community Association (OCA)",
```

sale_commission_product_criteria/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "contributors": ["Ilyas"],
```
pedrobaeza: This key is not used. Remove it.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,392 @@
...
+            return commission_item.fixed_amount
+        elif commission_item.commission_type == "percentage":
+            return subtotal * (commission_item.percent_amount / 100.0)
+
+
+class SaleCommission(models.Model):
+    _inherit = "sale.commission"
+
+    commission_type = fields.Selection(
+        selection_add=[("rules", "Rules")], ondelete={"rules": "cascade"}
```
pedrobaeza: Put a better name and don't delete on cascade (it will provoke problems on existing invoice lines):

```suggestion
        selection_add=[("product", "Product criteria")], ondelete={"rules": "set default"}
```

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,392 @@
...
+    sequence = fields.Integer(
+        "Sequence", default=1, help="The first in the sequence is the default one."
+    )
+
+    def write(self, vals):
+        res = super(SaleCommission, self).write(vals)
+        self.sort_items()
+        return res
+
+    def sort_items(self):
```
pedrobaeza: Why not putting this through `_order` in the model?

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,392 @@
...
+        # Regular with price list > Regular without price list
+        for com_type, seq in COMMISSION_ITEM_TYPES.items():
+            all_lines = self.item_ids.filtered(lambda x: x.applied_on == com_type)
+
+            sol_lines = all_lines.filtered(lambda x: x.based_on == "sol")
+            sol_lines.update({"sequence": seq + 100})
+
+    def action_archive(self):
+        res = super().action_archive()
+        self.item_ids.write({"active": False})
```
pedrobaeza: Can you avoid this and simply in the SQL don't take into account inactive commission types?

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,392 @@
...
+            raise ValidationError(
+                _(
+                    "There is posted Account Move Lines with this commission. "
+                    "Commission type change is not allowed."
+                )
+            )
+
+    def copy(self, default=None):
+        res = super().copy(default)
+        for cti in self.item_ids:
```
pedrobaeza: You don't need to do this. Just put `copy=True` in the one2many definition.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,392 @@
...
+COMMISSION_ITEM_TYPES = {
+    "0_product_variant": 1000,
+    "1_product": 10000,
+    "2_product_category": 100000,
+    "3_global": 1000000,
+}
+
+
+class SaleCommissionLineMixin(models.AbstractModel):
+    _inherit = "sale.commission.line.mixin"  # OCA sale_commission
```
pedrobaeza: Don't mix this model in the same file, as they are not directly related by a o2m relation.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,392 @@
...
+                raise ValidationError(
+                    _(
+                        "Please specify the product variant for "
+                        "which this rule should be applied"
+                    )
+                )
+
+    @api.onchange("commission_type")
+    def _onchange_commission_type(self):
+        if self.commission_type != "fixed":
```
pedrobaeza: This shouldn't go into this module.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,392 @@
...
+            if (
+                item.product_id
+                and item.product_id.product_tmpl_id != item.product_tmpl_id
+            ):
+                item.product_id = None
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for values in vals_list:
+            if values.get("applied_on", False):
```
pedrobaeza: Extract this code into a method, as it's used 2 times.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,392 @@
...
+            elif applied_on == "2_product_category":
+                values.update(dict(product_id=None, product_tmpl_id=None))
+            elif applied_on == "1_product":
+                values.update(dict(product_id=None, categ_id=None))
+            elif applied_on == "0_product_variant":
+                values.update(dict(categ_id=None))
+        res = super(CommissionItem, self).write(values)
+        # When the commission changes we need the product.template price
+        # to be invalided and recomputed.
+        self.flush()
```
pedrobaeza: This is done by Odoo, so it's not needed.

francesco-ooops:
Thanks a lot Pedro!

sale_commission_product_criteria/models/commission.py
ilyasProgrammer: it will be inherited in next modules and logic here will be more compex.

sale_commission_product_criteria/models/commission.py
ilyasProgrammer: Can you please tell how ? I dont get it.

ilyasProgrammer:
@pedrobaeza Thanks for the review. Updated. Please check some comments.

francesco-ooops:
@pedrobaeza any idea on what's wrong with the test?

pedrobaeza:
Seems a temporary glitch. Restarted.

sale_commission_product_criteria/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,16 @@
+This module allows to set in the same Commission Type different commission rates according to the product on SO/invoice line.
+
+This is made possible since this module adds a new "Rules" type to Commission Type and applies commission rates with the same logic of sale pricelist items.
```
francesco-ooops: Change "Rules" with "Product criteria"

sale_commission_product_criteria/readme/USAGE.rst
```diff
@@ -0,0 +1,7 @@
+To use features of this module, you need to:
+
+#. Go to Sales > Commission Management > Commission Types.
+#. Create a Commission Type with type = "Rules".
```
francesco-ooops: "Product criteria" here too

francesco-ooops (commented):
@ilyasProgrammer

sale_commission_product_criteria/models/commission.py
francesco-ooops: @pedrobaeza can you clarify?

sale_commission_product_criteria/models/commission.py
pedrobaeza: In this SQL:

https://github.com/OCA/commission/pull/414/files#diff-ae96df1a277be52a5924c783ffd66ee5aa73b373f4b090f66770d639139d7ba8R36

add the criteria for commission_id.active = True.

By the way, you are not filtering right now by item.active either.

ilyasProgrammer: ok got it. updated.

francesco-ooops:
@pedrobaeza good to be merged?

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,299 @@
...
+        ondelete={"product": "set default"},
+    )
+    item_ids = fields.One2many("commission.item", "commission_id", copy=True)
+    sequence = fields.Integer(
+        "Sequence", default=1, help="The first in the sequence is the default one."
+    )
+
+    def write(self, vals):
+        res = super(SaleCommission, self).write(vals)
+        self.sort_items()
```
pedrobaeza: Instead of this, order directly by the selection field.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,299 @@
...
+    )
+    categ_id = fields.Many2one(
+        "product.category",
+        "Product Category",
+        ondelete="cascade",
+        help="Specify a product category if this rule only applies to "
+        "products belonging to this category or its children categories. "
+        "Keep empty otherwise.",
+    )
+    based_on = fields.Selection(
```
pedrobaeza, line 102: I don't get the need of this.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,299 @@
...
+        values = self.validate_values(values)
+        return super(CommissionItem, self).create(values)
+
+    def write(self, values):
+        values = self.validate_values(values)
+        res = super(CommissionItem, self).write(values)
+        self.invalidate_cache()
+        return res
+
+    def validate_values(self, values):
```
pedrobaeza, line 253: Why not doing this in a method api.constrains?

sale_commission_product_criteria/models/sale.py
```diff
@@ -0,0 +1,52 @@
+# © 2023 ooops404
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+from odoo import api, fields, models
+
+
+class SaleOrderLineAgent(models.Model):
+    _inherit = "sale.order.line.agent"
+
+    discount = fields.Float(related="object_id.discount")
+    applied_commission_item_id = fields.Many2one("commission.item")
```
pedrobaeza, line 10: Do you need all these related fields?

sale_commission_product_criteria/models/commission.py
ilyasProgrammer: This method will be inherited in further modules:
https://github.com/OCA/commission/pull/421/files#diff-db23ccfdfe6db97ef62e546b8909cb7e32eeb9e56e8f3087d0670decc07f83a2R36

sale_commission_product_criteria/models/commission.py
ilyasProgrammer: Required for sorting commission items. More types will be added in other modules.

sale_commission_product_criteria/models/commission.py
ilyasProgrammer: In this method there is data modifications and  api.constrains designed only for checks. Or am i wrong ?

sale_commission_product_criteria/models/sale.py
ilyasProgrammer: Yes. discount is needed for calculations in sale_commission_product_criteria_discount. 
applied_commission_item_id - just to show which commission item was used.

ilyasProgrammer:
@pedrobaeza could you check again please

francesco-ooops:
Hi @pedrobaeza , this module goes hand in hand with the one in PR #421 , which extends it as follows: 

> This module extends sale_commission_product_criteria to allow user to restrict the commission granted to agent on a Sale Order Line according to the discount applied on the same line.
> 
> This is useful to implement a common commission structure for agents which provides a higher commission to agents granting a lower discount to customer, decreasing the commission proportionally as discount increases.

So our end goal is to provide to customer the possibility to grant commission according to discount on SOL. 

We have been trying our best to make this PR ready for merge. If there isn't any major blocking point, could we proceed with merge so we can finalize also the other PR? 

Thank you!

sale_commission_product_criteria/models/commission.py
pedrobaeza: You can do the same adding a value on the selection field. Doing this is counter-productive in performance terms. If you have to put values in between, name the selection options `10_`, `20_` instead of `1_`, `2_`, and put the values in between like `15_` or similar.

ilyasProgrammer: @pedrobaeza Perhaps this code should be implemented on commission.item level to improve performance a bit. But i dont understand how selection can help here. Quantity of commission items can be unpredicatable big. Why is it wrong to use sequence for sorting, which is standard odoo approach?

pedrobaeza: The problem is to sort on each write/create. This can also create loops very difficult to debug. You can sort by any needed field in the model. If quantity is a criteria, then sort by it.

ilyasProgrammer: ok. it has been redone to use only **_order**.

sale_commission_product_criteria/models/commission.py
pedrobaeza: I still don't get functionally the need of this.

sale_commission_product_criteria/models/commission.py
pedrobaeza: This is not needed according the latest SQL.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,275 @@
...
+
+class SaleCommission(models.Model):
+    _inherit = "sale.commission"
+
+    commission_type = fields.Selection(
+        selection_add=[("product", "Product criteria")],
+        ondelete={"product": "set default"},
+    )
+    item_ids = fields.One2many("commission.item", "commission_id", copy=True)
+    sequence = fields.Integer(
```
pedrobaeza: This sequence should go out as alien feature.

sale_commission_product_criteria/models/commission.py
ilyasProgrammer: https://github.com/OCA/commission/pull/421/files#diff-db23ccfdfe6db97ef62e546b8909cb7e32eeb9e56e8f3087d0670decc07f83a2R17
check here

sale_commission_product_criteria/models/commission.py
ilyasProgrammer: updated

sale_commission_product_criteria/models/commission.py
pedrobaeza: OK, seems a bit awkward, but not being more picky.

pedrobaeza (commented):
/ocabot merge nobump

francesco-ooops:
Thanks a lot @pedrobaeza for the guidance!

PR #415, opened 2023-04-17
[16.0][FIX] account_commission: button typo and period string

jguenat:
- cancel button string typo
- period string for invoice line name was "('\nPeriod: from 04/01/2023 to 04/30/2023',)"

### DIFF
--- a/account_commission/models/commission_settlement.py
+++ b/account_commission/models/commission_settlement.py
@@ -107,13 +107,12 @@ class CommissionSettlement(models.Model):
                         "product_id": product.id,
                         "quantity": -1 if settlement.total < 0 else 1,
                         "price_unit": abs(settlement.total),
-                        "name": (
-                            "\n"
-                            + _(
-                                "Period: from %(date_from)s to %(date_to)s",
-                                date_from=date_from.strftime(lang.date_format),
-                                date_to=date_to.strftime(lang.date_format),
-                            ),
+                        "name": product.with_context(lang=lang.code).display_name
+                        + "\n"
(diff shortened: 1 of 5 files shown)


pedrobaeza (changes requested):
Both were correct:

- The carriage return is on purpose and should be kept.
- The button text `_` is not a typo, but for selecting the letter to be the shortcut. Maybe this is not working anymore, but you should change it across all the translations.

jguenat:
I investigate older versions and the /n was to put he period string after (new line) the product description.
The _ in button string is indeed not used anymore, I updated with the current standard.
I took the liberty to use "z" like other cancel buttons and to remove the "_" in the Japanese translation.

account_commission/models/commission_settlement.py
```diff
@@ -107,13 +107,12 @@ def _prepare_invoice(self, journal, product, date=False):
...
                         "quantity": -1 if settlement.total < 0 else 1,
                         "price_unit": abs(settlement.total),
-                        "name": (
-                            "\n"
-                            + _(
-                                "Period: from %(date_from)s to %(date_to)s",
-                                date_from=date_from.strftime(lang.date_format),
-                                date_to=date_to.strftime(lang.date_format),
-                            ),
+                        "name": product.name
```
pedrobaeza:
```suggestion
                        "name": product.display_name
```

jguenat: I also added translation support :)

pedrobaeza (approved):

PR #416, opened 2023-04-17
[16.0][FIX] account_commission: revert settlement state to settled when deleting draft settlement invoice

jguenat:
- if you delete a draft settlement invoice, the settlement is stuck on "invoiced" state

### DIFF
--- a/account_commission/models/account_move.py
+++ b/account_commission/models/account_move.py
@@ -114,6 +114,13 @@ class AccountMove(models.Model):
                 res["arch"] = etree.tostring(invoice_xml)
         return res
 
+    def unlink(self):
+        """Put 'invoiced' settlements associated to the invoices back in settled state."""
+        self.invoice_line_ids.settlement_id.filtered(
+            lambda s: s.state == "invoiced"
+        ).write({"state": "settled"})
+        return super().unlink()
+
 
 class AccountMoveLine(models.Model):
     _inherit = [
--- a/account_commission/tests/test_account_commission.py
+++ b/account_commission/tests/test_account_commission.py
@@ -454,3 +454,14 @@ class TestAccountCommission(TestCommissionBase):
         settlements.make_invoices(self.journal, self.commission_product, grouped=True)
         invoices = settlements.mapped("invoice_id")
         self.assertEqual(2, invoices.settlement_count)
+
+    def test_unlink_settlement_invoice(self):
+        settlements = self._create_multi_settlements()
+        invoices = settlements.make_invoices(self.journal, self.commission_product)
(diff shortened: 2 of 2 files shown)


account_commission/models/account_move.py
```diff
@@ -114,6 +114,11 @@ def fields_view_get(
                 res["arch"] = etree.tostring(invoice_xml)
         return res
 
+    def unlink(self):
+        for invoice in self:
+            invoice.invoice_line_ids.settlement_id.state = "settled"
```
pedrobaeza: Put docstring here about why is the overriding, and you can do it in one step:

```suggestion
        self.invoice_line_ids.settlement_id.write({"state": "settled"})
```

jguenat: done :)

pedrobaeza:
Seeing the commit history, as I remember some tweaking on this, there's not too much sense in removing directly the invoice, and you have to pass through the cancel button, and there's already code for putting in a special state for this:

https://github.com/OCA/commission/commit/49b0ff3daca47d133ec3bd9317cb95db81456bdd#diff-c9164979c9473b61249db43c11ccb7ec045fee4e548c0f327dc636a8d161a15dR59

Reverting back as if nothing happened seems incorrect.

jguenat:
> Seeing the commit history, as I remember some tweaking on this, there's not too much sense in removing directly the invoice, and you have to pass through the cancel button, and there's already code for putting in a special state for this:
> 
> [49b0ff3#diff-c9164979c9473b61249db43c11ccb7ec045fee4e548c0f327dc636a8d161a15dR59](https://github.com/OCA/commission/commit/49b0ff3daca47d133ec3bd9317cb95db81456bdd#diff-c9164979c9473b61249db43c11ccb7ec045fee4e548c0f327dc636a8d161a15dR59)
> 
> Reverting back as if nothing happened seems incorrect.

Let's imagine someone use the wizard to create settlement invoices by mistake (wrong date input or forget to tick the "grouped" feature) and wants to cancel. I think deleting the draft invoice(s) is the more intuitive go to action.
Even if not intuitive someone can and will do it.
We could put the settlement state in "Invoice exception" like with the button_cancel but it would then require a lot of fiddling to correct a mistake potentially impacting a lot of settlements. You can really create tons of invoices with the wizard if you have 30+ agents.
What do you think?

pedrobaeza:
Please do it only if the previous state is `invoiced`, but not when on invoice exception.

pedrobaeza (approved):

pedrobaeza:
Check pre-commit

PR #418, opened 2023-04-26
[16.0][IMP] commission: Default commission_free instead of related

hildickethan:
Error case:
- Do normal settlement process for any product (sell it, invoice it, settle it)
- Change "Free of commission" on that same product
image

This is because it triggers computes through depends, reaching [this line](https://github.com/OCA/commission/blob/15.0/account_commission/models/account_move.py#L151) which causes the error (which is arguably probably suboptimal, but besides the point I think)

To me logically, old sale and invoice lines with commissions shouldn't be touched if product is no longer commissionable. I imagine this is what the `store=True` part of the related line level field `commission_free` is for, but it still changes the value if changes on the product.
To fix this I changed the `related` to a default value instead, this way it wouldn't change if the product is changed in the future. No data is lost since the previous field was already stored

Also changed the depends to the product directly in account_commission so it would use the line's field, making things more cohesive

### DIFF
--- a/commission/models/commission_mixin.py
+++ b/commission/models/commission_mixin.py
@@ -23,7 +23,7 @@ class CommissionMixin(models.AbstractModel):
     product_id = fields.Many2one(comodel_name="product.product", string="Product")
     commission_free = fields.Boolean(
         string="Comm. free",
-        related="product_id.commission_free",
+        compute="_compute_commission_free",
         store=True,
         readonly=True,
     )
@@ -49,6 +49,12 @@ class CommissionMixin(models.AbstractModel):
     def _compute_agent_ids(self):
         """Empty method that needs to be implemented in children models."""
         raise NotImplementedError()
+
+    @api.depends("product_id")
+    def _compute_commission_free(self):
+        """Compute instead of a simple related to have a proper initialized value."""
+        for line in self:
+            line.commission_free = line.product_id.commission_free
 
     @api.depends("commission_free", "agent_ids")
     def _compute_commission_status(self):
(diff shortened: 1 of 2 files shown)


commission/models/commission_mixin.py
```diff
@@ -23,8 +23,7 @@ class CommissionMixin(models.AbstractModel):
     product_id = fields.Many2one(comodel_name="product.product", string="Product")
     commission_free = fields.Boolean(
         string="Comm. free",
-        related="product_id.commission_free",
-        store=True,
+        default=lambda m: m.product_id.commission_free,
```
pedrobaeza: This is not valid, and it may work by chance. Change to a computed field if the related is not initialized on time.

hildickethan: Changed to compute

commission/models/commission_mixin.py
```diff
@@ -50,6 +50,11 @@ def _compute_agent_ids(self):
         """Empty method that needs to be implemented in children models."""
         raise NotImplementedError()
 
+    @api.depends("product_id")
+    def _compute_commission_free(self):
```
pedrobaeza, line 54: It's advisable to comment/docstring decisions that are difficult to understand just reading code:

```suggestion
    def _compute_commission_free(self):
        """Put as computed instead of a simple related
        to have a proper initialized value.
        """
```

pedrobaeza (approved):

commission/models/commission_mixin.py
hildickethan: @pedrobaeza Done 🫡

pedrobaeza (approved):

jguenat (approved):
LG thanks

PR #422, opened 2023-05-03
[14.0] [ADD] sale_commission_check_deposit

odooNextev:
Add a new module to manage checks and commission
Add safety days on model and view


stefano-ooops (approved):
Functional OK

matteotognini (approved):
Functional ok

sale_commission_check_deposit/views/account_journal_view.xml
```diff
@@ -0,0 +1,17 @@
...
+    <record model="ir.ui.view" id="account_journal_check">
+        <field name="name">account.journal.check</field>
+        <field name="model">account.journal</field>
+        <field name="inherit_id" ref="account.view_account_journal_form" />
+        <field name="arch" type="xml">
+            <field name="type" position="after">
+                <field name="is_check_journal" />
+                <field
+                    name="safety_days"
+                    attrs="{'invisible': [('is_check_journal', '=', False)]}"
```
GSLabIt: maybe need to be `required` if `is_check_journal`, isn't it?

sale_commission_check_deposit/wizard/wizard_check_settle.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _get_agent_lines(self, agent, date_to_agent):
+        """
+        Exclude invoices with checks payment if at least safety days
+        haven't passed since expiration date set on check deposit.
+        """
+        agent_lines = super()._get_agent_lines(agent, date_to_agent)
+        for line in agent_lines:
+            if (
+                line.commission_id.invoice_state == "paid"
```
GSLabIt: can u filter directly from agent_lines instead of looping through it? kinda:

```
agent_lines.filtered(lambda r: r.commision_id.invoice_state == "paid" and line.invoice_id.invoice_payments_widget != "false")
```

sale_commission_check_deposit/wizard/wizard_check_settle.py
```diff
@@ -0,0 +1,47 @@
...
+                    check_deposit_line = move.line_ids.filtered(
+                        lambda x: x.check_deposit_id
+                    )
+                    if not check_deposit_line or (
+                        check_deposit_line.date_maturity
+                        + timedelta(days=check_deposit_line.journal_id.safety_days)
+                        > date.today()
+                    ):
+                        agent_lines -= line
+                        break
```
GSLabIt: why `break`? is not possible to have more than 1 line to remove?

sale_commission_check_deposit/wizard/wizard_check_settle.py
```diff
@@ -0,0 +1,47 @@
...
+    def get_move_ids(self, payments_widget):
+        move_dict = json.loads(payments_widget)
+        account_move_ids = [content["move_id"] for content in move_dict["content"]]
+        move_lines = self.env["account.move.line"].search(
+            [
+                ("move_id", "in", account_move_ids),
+                ("journal_id.is_check_journal", "=", True),
+            ]
+        )
+        return move_lines.mapped("move_id")
```
GSLabIt: may be would be possible to return account move lines here, kinda:

```
move_lines.filtered(lambda x: x.check_deposit_id)
```

why returning `move_id`?

GSLabIt (changes requested):

sale_commission_check_deposit/wizard/wizard_check_settle.py
odooNextev: If function returns only move_lines with check_deposit_id it will not remove agent line without it.

GSLabIt: if needed, you can always ref by `move_line.move_id`, isn't it?

GSLabIt: or return `move_lines.filtered(lambda x: x.check_deposit_id).mapped("move_id")`

GSLabIt: or even: 
```
move_lines.filtered(lambda x: x.check_deposit_id and x.check_deposit_line.date_maturity
                        + timedelta(days=x.check_deposit_line.journal_id.safety_days)
                        < date.today()).mapped("move_id")
```

so that u return only move having required conditions

odooNextev: Yes, but I also need move_id without check_deposit_id because I need to remove that agent_line. If I filter to get only move_lines with check_deposit_id I can't remove that agent_lines.

GSLabIt: sorry not getting the point .. u have all lines and all lines that match conditions, what about removing all at once like` all_lines - all lines that match conditions = left lines to remove`. Am I missing something?  

Other than that u are removing 1 line and than breaking loop .. 

surely i'm missing something, lets wait some other review

odooNextev: Sorry I understand only now what you mean. I will correct now. Thanks

sale_commission_check_deposit/views/account_journal_view.xml
odooNextev: Trying to save safety_days without data, it will save it with 0. Do you think it is ok?

GSLabIt: i think is better to have it required  when `is_check_journal` is set

sale_commission_check_deposit/wizard/wizard_check_settle.py
```diff
@@ -0,0 +1,40 @@
...
+        Exclude invoices with checks payment if at least safety days
+        haven't passed since expiration date set on check deposit.
+        """
+        agent_lines = super()._get_agent_lines(agent, date_to_agent)
+        for line in agent_lines.filtered(
+            lambda r: r.commission_id.invoice_state == "paid"
+            and r.invoice_id.invoice_payments_widget != "false"
+        ):
+            if self.get_move_ids(line.invoice_id.invoice_payments_widget):
+                agent_lines -= line
```
GSLabIt: not getting this one, can you elaborate?

odooNextev: Inside get_move_ids I return the difference between all move_id that have move_lines that are in check journal and all move_id that have lines with check_deposit_id and safety_days are passed from date.today. If that not return any move_id it means that I need to settle all agent_lines, otherwise I need to remove that specific agent_line.

GSLabIt: what about:

```
     move_lines = self.env["account.move.line"].search(
            [
                ("move_id", "in", account_move_ids),
                ("journal_id.is_check_journal", "=", True),
                ("check_deposit_id", "=", True),
            ])
     return any([x.date_maturity + timedelta(days=x.journal_id.safety_days) < date.today() for x in move_lines])
```

It check if you have at least 1 `move_line`, with `check_deposit_id`, having `maturity + safety < today` and return `true`. In this case you can remove line from agent.

Could this do the job?

odooNextev: If returns at least 1 move_line it means that I need to settle that commission. But if it returns nothing I will remove it. I'll publish it soon. Thanks.

GSLabIt: > If returns at least 1 move_line it means that I need to settle that commission. But if it returns nothing I will remove it. I'll publish it soon. Thanks.

exactly the contrary

odooNextev: I published the changes..

sale_commission_check_deposit/wizard/wizard_check_settle.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        Exclude invoices with checks payment if at least safety days
+        haven't passed since expiration date set on check deposit.
+        """
+        agent_lines = super()._get_agent_lines(agent, date_to_agent)
+        for line in agent_lines.filtered(
+            lambda r: r.commission_id.invoice_state == "paid"
+            and r.invoice_id.invoice_payments_widget != "false"
+        ):
+            if not self.get_move_ids(line.invoice_id.invoice_payments_widget):
```
GSLabIt, line 23: why `not`here? 

the condition is: `x.date_maturity + timedelta(days=x.journal_id.safety_days) < date.today()`.

The function return `True` if there is at least 1 move_line match condition.  

As far as i understand if condition is `True` you should remove `agent_lines`, am i missing something?

odooNextev: I need to remove agent lines if it return False because if it match that conditions it means that the check is correctly deposited and safety days are passed. If not it means that I'm consider the invoice not payed yet and so I need to remove agent line.

GSLabIt: than use `all` in place of `any` so that will be `True` only if check is correctly deposited and all the lines passed safety days

GSLabIt (approved):

francesco-ooops:
@pedrobaeza does this look good to you?

pedrobaeza:
Merging based on reviews:

/ocabot merge nobump

Please put as maintainer of the modules for being self-independent on patches over this module.

PR #431, opened 2023-05-30
[16.0] [IMP] commission: Settlement mail thread activity and report

jguenat:
Add mail & activity support for Settlement

Improve the existing document report:
- external_layout instead of internal_layout
- language support
- remove the Invoice column which is related to the settlement invoice ; it might not exist and I find it strange to be next to the Invoice line column which is not related (it's the commission source). 
The report changes might be too opinionated so It's on a separate commit I could remove based on feedback.

Idea is to use settlement as a communication channel with agents, who might be external.

### DIFF
--- a/account_commission/views/report_settlement_templates.xml
+++ b/account_commission/views/report_settlement_templates.xml
@@ -1,21 +1,36 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <template id="report_settlement" inherit_id="commission.report_settlement">
-        <xpath expr="//th" position="before">
-            <th>Invoice date</th>
-            <th>Invoice</th>
-            <th>Invoice line</th>
+    <template
+        id="report_settlement_document"
+        inherit_id="commission.report_settlement_document"
+    >
+        <xpath expr="//div[@id='informations']" position="inside">
+            <div class="col-auto col-3 mw-100 mb-2" t-if="o.invoice_id">
+                <strong>Settlement invoice:</strong>
+                <p t-field="o.invoice_id" />
+            </div>
         </xpath>
-        <xpath expr="//table/tbody/tr/td" position="before">
-            <td>
-                <span t-field="l.date" />
-            </td>
-            <td>
-                <span t-field="l.settlement_id.invoice_id" />
-            </td>
-            <td>
-                <span t-field="l.invoice_line_id" />
-            </td>
+        <xpath expr="//th" position="after">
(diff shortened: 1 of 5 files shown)


commission/models/commission_settlement.py
```diff
@@ -79,6 +80,15 @@ def _compute_can_edit(self):
     def action_cancel(self):
         self.write({"state": "cancel"})
 
+    def _message_get_suggested_recipients(self):
```
pedrobaeza: Why not doing the agent auto-subscribed on creation instead? Similar to what happens with the customer when confirming a sales order

pedrobaeza:
Better to split in 2 PRs for the report one not to block the other, and to include an screenshot of the before and after of the report for visually comparing.

commission/models/commission_settlement.py
jguenat: yes I agree it's more in line with standard odoo behaviour

jguenat:
I remember now why it is easier to modify both reports at the same time because account_commission depends on the commission one.

Should I spend time making it work with two separte PR or going back as a bundle ? :)

Anyways here is already the difference in the base report:

Before
commission_before

After
commission_after

I added the date because I think it's better to have it already in the base report instead of adding it in account_commission.

pedrobaeza:
OK, then let's keep it this way, but please do CI happy.

jguenat:
Improvement of account_commission report.

Before:
account_commission_before

After:
account_commission_after

I removed the "invoice" column, which is the settlement invoice and not the source commission invoice. I think it added confusion.

account_commission/views/report_settlement_templates.xml
```diff
@@ -1,21 +1,24 @@
...
-    <template id="report_settlement" inherit_id="commission.report_settlement">
-        <xpath expr="//th" position="before">
-            <th>Invoice date</th>
-            <th>Invoice</th>
-            <th>Invoice line</th>
+    <template
+        id="report_settlement_document"
+        inherit_id="commission.report_settlement_document"
+    >
+        <xpath expr="//th" position="after">
```
pedrobaeza, line 13: These changes should act only if the settlement type is from invoices.

jguenat: done, also (re)added the settlement invoice but in the information area:
account_commission_after_with_settlement_invoice

pedrobaeza (approved):

pedrobaeza:
Let's merge this welcome improvement:

/ocabot merge minor

PR #440, opened 2023-07-14
[15.0][IMP] account_commission: skip not invoice lines

cormaza:
- Modified original module account_commission to make filter on applicable lines heritable and skip not invoice lines

### DIFF
--- a/account_commission/models/account_move.py
+++ b/account_commission/models/account_move.py
@@ -136,6 +136,13 @@ class AccountMoveLine(models.Model):
         for record in self:
             record.any_settled = any(record.mapped("agent_ids.settled"))
 
+    def _filter_commission_applicable_lines(self):
+        return self.filtered(
+            lambda x: x.move_id.partner_id
+            and x.move_id.move_type[:3] == "out"
+            and not x.exclude_from_invoice_tab
+        )
+
     @api.depends("move_id.partner_id")
     def _compute_agent_ids(self):
         for res in self:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Wouldn't be enough to only filter those lines with `exclude_from_invoice_tab=False` for removing also anglo-saxon lines?

cormaza:
> Wouldn't be enough to only filter those lines with `exclude_from_invoice_tab=False` for removing also anglo-saxon lines?

You're right, but I think it's also a good idea to make it an independent function that returns the lines, to be able to customize, what do you think?

pedrobaeza:
OK, extract it to the hook with such filter, and your module is not needed at all.

cormaza:
> OK, extract it to the hook with such filter, and your module is not needed at all.

What do you suggest to continue in this pr or open another?

pedrobaeza:
What is easier for you. If continuing in this PR, rename the commit and PR title.

account_commission/models/account_move.py
```diff
@@ -136,6 +136,13 @@ def _compute_any_settled(self):
         for record in self:
             record.any_settled = any(record.mapped("agent_ids.settled"))
 
+    def filter_commission_applicable_lines(self):
```
pedrobaeza: Make it private, as it's not called from RPC:

```suggestion
    def _filter_commission_applicable_lines(self):
```

cormaza: Done!

pedrobaeza (approved):

pedrobaeza:
Let's clean the pending PRs:

/ocabot merge patch

Please fw-port it to v16

PR #441, opened 2023-07-28
[14 0] [IMP] sale_commission: possibility to add payment date limit on settle commissions

odooNextev:
This module add a field on wizard that allow to restring commission settle base on payment date. All invoice that are paid after that date will be removed from settle

### DIFF
--- a/sale_commission/wizard/wizard_settle.py
+++ b/sale_commission/wizard/wizard_settle.py
@@ -1,6 +1,6 @@
 # Copyright 2014-2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from datetime import date, timedelta
+from datetime import date
 
 from dateutil.relativedelta import relativedelta
 
@@ -11,7 +11,20 @@ class SaleCommissionMakeSettle(models.TransientModel):
     _name = "sale.commission.make.settle"
     _description = "Wizard for settling commissions in invoices"
 
-    date_to = fields.Date("Up to", required=True, default=fields.Date.today())
+    date_to = fields.Date(
+        "Invoice date up to",
+        help="For invoice-based commissions, settlements will be created for invoices \
+            with date up to the one set in this field.",
+        required=True,
+        default=fields.Date.today(),
+    )
+    date_payment_to = fields.Date(
+        "Payment date up to",
+        help="For payment-based commissions, settlements will be created for payments \
+            with date up to the one set in this field.",
+        default=fields.Date.today(),
+    )
+
     agent_ids = fields.Many2many(
         comodel_name="res.partner", domain="[('agent', '=', True)]"
     )
@@ -99,12 +112,16 @@ class SaleCommissionMakeSettle(models.TransientModel):
                     pos += 1
                     if line._skip_settlement():
                         continue
+                    if self.date_payment_to and line._skip_future_payments(
+                        self.date_payment_to
+                    ):
+                        continue
                     if line.invoice_date > sett_to:
                         sett_from = self._get_period_start(agent, line.invoice_date)
                         sett_to = self._get_next_period_date(
                             agent,
                             sett_from,
-                        ) - timedelta(days=1)
+                        ) - relativedelta(days=1)
                         settlement = self._get_settlement(
                             agent, company, sett_from, sett_to
                         )
--- a/sale_commission/models/account_move.py
+++ b/sale_commission/models/account_move.py
@@ -221,3 +221,16 @@ class AccountInvoiceLineAgent(models.Model):
             self.commission_id.invoice_state == "paid"
             and self.invoice_id.payment_state not in ["in_payment", "paid"]
         ) or self.invoice_id.state != "posted"
+
+    def _skip_future_payments(self, date_payment_to):
+        if self.commission_id.invoice_state == "paid":
+            payments_dates = []
+            for (
+                _partial,
+                _amount,
+                counterpart_line,
+            ) in self.invoice_id._get_reconciled_invoices_partials():
+                payments_dates.append(counterpart_line.date)
+            if any(date_payment_to < date for date in payments_dates):
+                return True
+        return False
--- a/sale_commission/readme/CONTRIBUTORS.rst
+++ b/sale_commission/readme/CONTRIBUTORS.rst
@@ -7,6 +7,8 @@
 * Oihane Crucelaegui <oihanecruce@gmail.com>
 * Nicola Malcontenti <nicola.malcontenti@agilebg.com>
 * Aitor Bouzas <aitor.bouzas@adaptivecity.com>
+* Nextev <web@nextev.it>
+
 
 * `Tecnativa <https://www.tecnativa.com>`__:
 
--- a/sale_commission/readme/USAGE.rst
+++ b/sale_commission/readme/USAGE.rst
@@ -38,6 +38,8 @@ For settling the commissions to agents:
    want to create commissions. It should be at least one day after the last
    period date. For example, if you settlements are monthly, you have to put
    at least the first day of the following month.
+#. You can select to settle commissions up to a specified payment date.
+   This only applies to commission types with "invoice status=payment based".
 #. You can settle only certain agents if you select them on the "Agents"
    section. Leave it empty for settling all.
 #. Click on "Make settlements" button.
--- a/sale_commission/tests/test_sale_commission.py
+++ b/sale_commission/tests/test_sale_commission.py
@@ -2,7 +2,8 @@
 # Copyright 2020 Tecnativa - Manuel Calero
 # License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
 
-import dateutil.relativedelta
+from datetime import datetime
+
 from dateutil.relativedelta import relativedelta
 
 from odoo import fields
@@ -83,6 +84,15 @@ class TestSaleCommission(SavepointCase):
                 "settlement": "monthly",
                 "lang": "en_US",
                 "commission_id": cls.commission_net_invoice.id,
+            }
+        )
+        cls.agent_monthly_paid = cls.res_partner_model.create(
+            {
+                "name": "Test Agent - Monthly Net Paid",
+                "agent": True,
+                "settlement": "monthly",
+                "lang": "en_US",
+                "commission_id": cls.commission_net_paid.id,
             }
         )
         cls.agent_biweekly = cls.res_partner_model.create(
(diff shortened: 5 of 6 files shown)


francesco-ooops (approved):
Functional ok!

sale_commission_settlement_payment_date/README.rst
HekkiMelody, line 1: You can generate the readme using `oca-gen-addon-readme`

https://github.com/OCA/maintainer-tools#readme-generator

sale_commission_settlement_payment_date/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "Commission Settlement base on payment date",
```
HekkiMelody:
```suggestion
    "name": "Commission Settlement based on payment date",
```

sale_commission_settlement_payment_date/static/description/index.html
HekkiMelody, line 1: This also need to be generated again (`oca-gen-addon-readme` will take care of this as well)

sale_commission_settlement_payment_date/wizard/wizard_settle.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def _get_agent_lines(self, agent, date_to_agent):
+        """
+        Exclude invoices with invoice payment date is after date_payment_to
+        """
+        agent_lines = super()._get_agent_lines(agent, date_to_agent)
+        for line in agent_lines.filtered(
+            lambda r: r.commission_id.invoice_state == "paid"
+            and self.date_payment_to != "false"
+        ):
```
HekkiMelody: `date_payment_to` will always be different from the string `"false"` since it can only be a Date object or the boolean `False`

In general, I don't think this particular check should go in the lambda since it's not meant to filter `agent_lines` but it's meant to check if the filtering should happen at all. Additionally I'd group the other checks together since they're not logically separate.

```python
if self.date_payment_to:
    for line in agent_lines.filtered(lambda r: r.commission_id.invoice_state == "paid" and r.invoice_id.invoice_payments_widget and not self.check_payment_date(r)):
```

or 

```python
if self.date_payment_to:
    for line in agent_lines:
        if line.commission_id.invoice_state == "paid" and line.invoice_id.invoice_payments_widget and not self.check_payment_date(line):
```

(and then let Black format that of course :) )

sale_commission_settlement_payment_date/wizard/wizard_settle.py
```diff
@@ -0,0 +1,39 @@
...
+            if (
+                line.invoice_id.invoice_payments_widget != "false"
```
HekkiMelody: Same as above. It's a text field. It'll probably be empty (`""` An empty string) or `False`, but never the string `"false"`. Might as well just do an `if line.invoice_id.invoice_payments_widget`

HekkiMelody (changes requested):

sale_commission_settlement_payment_date/wizard/wizard_settle.py
odooNextev: @aleuffre Thanks for your review. I know that sound weird, but invoice_payments_widget it is possible to be a string "false".

HekkiMelody: ok, that's interesting haha

sale_commission_settlement_payment_date/wizard/wizard_settle.py
odooNextev: Here condition on "false" it's my mistake. I will change it and re-publish all changes

sale_commission_settlement_payment_date/wizard/wizard_settle.py
odooNextev: It happens when an invoice is not paid yet. I could change condition on invoice.state without checking invoice_payments_widget. What could you recommend me?

sale_commission_settlement_payment_date/tests/test_commission_payment_date.py
```diff
@@ -0,0 +1,167 @@
...
+            [
+                (
+                    "agent_id",
+                    "=",
+                    self.agent_monthly.id,
+                ),
+                ("state", "=", "settled"),
+            ]
+        )
+        self.assertEqual(1, len(settlements))
```
HekkiMelody: I'm not entirely sure, so please correct me if I'm wrong, but it seems to me that this test only creates a single payment and checks that it's included in the wizard.

You could improve the test by creating two payments (or two sales orders with one payment each, if it's easier), one in the past (as the one you've already created) and one in the future, and making sure that when you call the wizard, only the one before `date_payment_to` is actually included in the wizard and the other is not.

As it is, if I understand correctly, your module could do nothing and this test would still pass.

HekkiMelody (commented):
Code looks good, not sure about the test

HekkiMelody (approved):
Code review, LGTM

HekkiMelody:
Only thing, try to make the first line of the commit message shorter! It should be less than 80 characters in total.

You can add more details in other lines.

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id48

francesco-ooops:
@pedrobaeza what do you think?

francesco-ooops:
@pedrobaeza good for merge?

pedrobaeza:
In this case, I consider reasonable to add this to the base module, but if you want to keep it as separate module, no problem.

francesco-ooops:
We'll gladly turn this PR into an IMP to sale_commission :)

@odooNextev can you proceed?

pedrobaeza:
Try to make it in one shot excluding from the initial domain instead of removing them later.

HekkiMelody (approved):
Generally code review LGTM.

There's some minor code cleanup that could be done (in one file, relativedelta is imported twice, with one import overwriting the other; also it's slightly confusing to be using both timedelta and relativedelta, especially when the latter is much nicer to use) but it doesn't really matter.

I'd echo the request by Pedro Baeza of trying to move as much of the search as possible to the domain in the `.search` method so it's done in the DB directly rather than in python, but I don't know how to do get the payment date info so I can't give advice there without looking into it in more detail.

HekkiMelody:
Oh, and as always, make sure to squash into a single commit, and follow OCA guidelines for the commit message

odooNextev:
Hi @aleuffre, thanks for your review. I made some cleanup that you recommended to me. I tried to move everything inside a single domain but it seems, in my opinion,  not possible.

sale_commission/wizard/wizard_settle.py
```diff
@@ -140,4 +143,21 @@ def _get_agent_lines(self, agent, date_to_agent):
             ],
             order="invoice_date",
         )
+        if self.date_payment_to:
+            aila = aila.filtered(
+                lambda line: line.commission_id.invoice_state != "paid"
+                or line.invoice_id.invoice_payments_widget == "false"
+                or self.check_payment_date(line)
+            )
```
HekkiMelody: If you want to limit lines to a certain payment date, aren't you with this including all lines that aren't paid as well?

odooNextev: I changed it and now I think that's works properly. Is it possible to call check_payment_date method inside domain?

HekkiMelody: unfortunately not. Domains inside a `.search` are translated into SQL and passed to the DB directly. It's also why you can't use fields that are computed and not stored, since they don't exist as columns in the DB.

sale_commission/wizard/wizard_settle.py
```diff
@@ -140,4 +143,23 @@ def _get_agent_lines(self, agent, date_to_agent):
...
+    def check_payment_date(self, agent_line):
+        move_dict_content = json.loads(agent_line.invoice_id.invoice_payments_widget)[
+            "content"
+        ]
+        return all(
+            [
+                self.date_payment_to > datetime.strptime(x["date"], "%Y-%m-%d").date()
+                for x in move_dict_content
+            ]
+        )
```
HekkiMelody: Done some digging into this in order to get around using the widget.

As far as I can tell there is no simple field that tells you when an invoice is fully paid. But you can get all the payments associated with an invoice (self.env["account.payment"].search("reconciled_invoice_ids", "in", invoice.id) and then check the date on those.

I also think that this type of check shouldn't be done here but higher up in this file (lines 103 and following).

There's also a function `_skip_settlement` that does part of the filtering you already want. So perhaps another `if condition: continue` up there, right after `_skip_settlement` would make the most sense.

HekkiMelody: Also, I'd like to point out that according to the existing logic of `_skip_settlement`, it seems to me (please correct me if I'm wrong) that as long as the invoice is partially paid, the commission settlement is created in full (for commissions of type "Payment Based") so I think we should respect that for now and maybe change the logic in a future PR

HekkiMelody (approved):
Looks clean, finally! :D

Just squash into a single commit following OCA conventions please.

francesco-ooops:
> Try to make it in one shot excluding from the initial domain instead of removing them later.

@pedrobaeza could it work like this?

sale_commission/wizard/wizard_settle.py
```diff
@@ -12,6 +12,8 @@ class SaleCommissionMakeSettle(models.TransientModel):
     _description = "Wizard for settling commissions in invoices"
 
     date_to = fields.Date("Up to", required=True, default=fields.Date.today())
+    date_payment_to = fields.Date("Date Payment Up to", default=fields.Date.today())
```
pedrobaeza: The name of the field and the label is not very intuitive. Also non having help doesn't help, hehe. Can you please look for other name and put a help for this new field.

francesco-ooops: @pedrobaeza since this is becoming a feature of base module, I would suggest changing labels of both date fields in the wizard for consistency: 

Invoice date up to
Payment date up to

shall we proceed?

pedrobaeza: OK

odooNextev: https://github.com/OCA/commission/pull/441/files#diff-7dfa7dd0fc23295fc2cd2da1c87e3535d33ddbf4480da661bdab6a3eaab7ea51R14

francesco-ooops: @pedrobaeza go?

pedrobaeza (approved):
Thanks for the changes

/ocabot merge minor

PR #442, opened 2023-08-02
[FIX][12.0]sale_commission: fix internal error trying to print int using %s

odooNextev:


odooNextev:
@pedrobaeza Could you please check this simple fix? Thanks

pedrobaeza:
But why is this needed? This was working on my side, converting the int to string by Python itself.

odooNextev:
On my side I had this error before this fix. Maybe it is because I was working with python 3.6.
/odoo/lib/python3.6/site-packages/odoo/addons/sale_commission/models/sale_commission_mixin.py", line 67, in _compute_commission_status
    ) % len(line.agents)
TypeError: not all arguments converted during string formatting

pedrobaeza:
Maybe it's a question of translations? Are you using it in English or any other language?

PR #444, opened 2023-08-22
[15.0][FIX] account_commission: revert settlement state to settled when deleting draft settlement invoice

jguenat:
backport of #416


pedrobaeza (approved):
And let's go with this one as well:

/ocabot merge patch

PR #405, opened 2023-03-15
[15.0][IMP] commission: multi-currency support

alexeirivera87:
Hello

Could someone please check these changes ?

regards

PR #446, opened 2023-08-22
[15.0][IMP] commission: multi-currency support

jguenat:
Supersede #405 

Improvements based on review in #405 are in a separate commit that will be fixed up later.

### DIFF
--- a/account_commission/models/account_move.py
+++ b/account_commission/models/account_move.py
@@ -211,6 +211,7 @@ class AccountInvoiceLineAgent(models.Model):
     currency_id = fields.Many2one(
         related="object_id.currency_id",
         readonly=True,
+        store=True,
     )
 
     @api.depends(
--- a/account_commission/models/account_move.py
+++ b/account_commission/models/account_move.py
@@ -210,8 +210,6 @@ class AccountInvoiceLineAgent(models.Model):
     )
     currency_id = fields.Many2one(
         related="object_id.currency_id",
-        readonly=True,
-        store=True,
     )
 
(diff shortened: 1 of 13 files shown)


pedrobaeza:
Shouldn't we have in settlement lines both amount in company currency and in the original currency?

jguenat:
If we need to settle the settlement with an invoice in the company currency it could be useful indeed.
We need to be careful about currency exchange rate. On the settlement line the rate from the source invoice date is used, then when we make the settlement invoice it will be another rate.
This could lead to confusion.

Maybe if we are able to choose the settlement invoice currency (original or company) then there is enough flexibility.

Next week I have a meeting with a customer that use commissions in different currency. I'll ask how they manage it actually. I guess it depends on the commission agreement.

JordiMForgeFlow:
Hi @jguenat 

Are there any plans to continue with this PR?

jguenat:
Hi @JordiMForgeFlow ,

I've been using this module for a client and it seems to work like this but the real commission on foreign currency will be settled early 2024 and maybe I'll have some adjustment to do then.

But if you have time to review this we can already go forward.

JordiMForgeFlow:
Hi @jguenat 

Any chance the PR could be rebased?

jguenat:
@JordiMForgeFlow I rebased but we can continue there #479 no problem

JordiMForgeFlow (approved):
Code + functional review LGTM!

pedrobaeza (approved):
/ocabot merge major

PR #447, opened 2023-08-25
[15.0][FIX] account_commission: redefine partner can cause issue

jguenat:
- partner is already defined above
- self can contain multiple records with different agents if overriding [grouping criteria](https://github.com/OCA/commission/blob/da5498012a2346b8e8e3a9bc7980820e48fe81fe/account_commission/models/commission_settlement.py#L120C9-L120C35)

### DIFF
--- a/account_commission/models/commission_settlement.py
+++ b/account_commission/models/commission_settlement.py
@@ -93,7 +93,7 @@ class CommissionSettlement(models.Model):
                 line_form.quantity = -1 if settlement.total < 0 else 1
                 line_form.price_unit = abs(settlement.total)
                 # Put period string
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
You should instead put `partner = settlement.agent_id`, as the intention of this is to put the line in the agent's regional conf. As you say, if the grouping criteria is another, there can be several agents involved, each one with a different language.

jguenat:
I though of that but I found it strange to have potentially a different language per invoice line.
If it was intentional then let's keep it.

pedrobaeza (approved):
/ocabot merge patch

PR #456, opened 2023-09-22
[14.0][FIX] sale_commission_geo_assign_product_domain: upd security and groups in view

HekkiMelody:
Fixes an error where users with permission: Sales: own documents could not access any contact.

Also makes the access rule more similar to other modules of this same repo.


francesco-ooops (approved):
Functional ok, you can merge!

francesco-ooops:
@pedrobaeza can you please merge as it's blocking? 

plus we forgot to add maintainer, now we did so we won't bother you on this one :)

PR #458, opened 2023-09-26
[15.0][FIX] sale_commission_salesman, fix singleton issue

cormaza:
Issue when function is called by a multiple records


pedrobaeza (approved):
Obvious fix.

/ocabot merge patch

PR #460, opened 2023-10-03
[FIX] sale_commission_agent_restrict: agent can create res.partner

HekkiMelody:
Quick fix for a bug that meant an agent could not create any contact because the form would pre-populate the value of payment terms.


francesco-ooops (approved):
Functional ok!

PR #463, opened 2023-10-13
 [16.0][FIX] account_commission: No skip reversed invoices in settlements

omar7r:
Hi,

If invoice reach 'reversed' value in payment_state and settlements are executed when invoice is paid, it doesn't appear in the settlement. Then, we could be settle the refund invoice and not the original invoice because is marked as reversed not paid.

Regards


pedrobaeza (approved):
Yes, totally!

/ocabot merge patch

Please backport it to v15/v14

PR #476, opened 2023-11-21
[15.0][FIX] account_commission: mandatory fields cannot be set to False.

angelgarciadelachica:
In the _compute_agent_ids function, two mandatory fields are set to False: `date` and `commission_id`.
https://github.com/OCA/commission/blob/c25ef03f47f12b3d34e6454eaae3f8cc3b19adbc/account_commission/models/account_move.py#L156-L164

An example to reproduce the error is:

1. Generate a sale and establish a customer that has a salesperson set up.
2. Add one or more products
3. Validate sale
4. Generate invoice
5. Validate and pay invoice
6. On the other hand from the commissions module we will go to the 'Settle Commissions' menu and we will set: 
> Up to'= current_date
> Settlement Type' = 'Sales invoice'.
> Add the Agent and click on 'Make settlements'.

7. Back to the invoice -> set it to draft. 
8. Change the customer and click on Save. 

We will see the following error:

Captura desde 2023-11-21 17-48-08

In this fix the mandatory fields are removed.

### DIFF
--- a/account_commission/models/account_move.py
+++ b/account_commission/models/account_move.py
@@ -152,16 +152,6 @@ class AccountMoveLine(models.Model):
 
     @api.depends("move_id.partner_id")
     def _compute_agent_ids(self):
-        for res in self:
-            settlement_lines = self.env["commission.settlement.line"].search(
-                [("invoice_line_id", "=", res.id)]
-            )
-            for line in settlement_lines:
-                line.date = False
-                line.agent_id = False
-                line.settled_amount = 0.00
-                line.currency_id = False
-                line.commission_id = False
         self.agent_ids = False  # for resetting previous agents
         for record in self._filter_commission_applicable_lines():
             if not record.commission_free and record.product_id:


angelgarciadelachica:
@HaraldPanten @ValentinVinagre

ValentinVinagre (approved):
LGTM 👍🏻

pedrobaeza (changes requested):
You should NEVER write in a compute. That triggers the DB write and that is not roll-backed if you discard the changes, among other ugly things.

Please indicate the steps to reproduce the problem with current OCA modules, and then, an alternative solution may be proposed.

angelgarciadelachica:
@pedrobaeza 
I have edited the description of the PR to add an example to reproduce the error with the runboat. 
On the other hand, in this case the `_compute_agent_ids` function recalculates the `agent_ids` when the customer of the invoice is changed, but `agent_ids` is not compute.

https://github.com/OCA/commission/blob/c25ef03f47f12b3d34e6454eaae3f8cc3b19adbc/account_commission/models/account_move.py#L132C5-L132C14

In this case it would also be incorrect?

THX!

pedrobaeza:
As said, a `write` in a compute method is always forbidden. Any other formula should be used, but not that one. I will check more later.

angelgarciadelachica:
Hi @pedrobaeza!
I have left it again with the for loop but removed the required fields. Could you check it?
THX!

angelgarciadelachica:
Hi @pedrobaeza
Have you been able to check it?
THX!

pedrobaeza:
Hi, I'm seeing that code, and it doesn't have any sense. I'm trying to trace back where it comes from, and in 14.0, it wasn't there: https://github.com/OCA/commission/blob/14.0/sale_commission/models/account_move.py#L115

so it was something added silently on the migration to 15.0. Please remove all that code, as a compute of this way must never change things outside its scope (it doesn't matter if with `write`, or with dot notation). @AungKoKoLin1997, it seems you introduce such code. Can you say why?

AungKoKoLin1997:
> it seems you introduce such code. Can you say why?

My intention was to clear the settlement lines that are generated when we change the partner_id in account.move. The agent_ids are set to false when we change the partner_id. So, I thought we should clear the settlement lines that are generated from the old agent_ids. But I think I forgot to check if this functionality is working well when I did the migration.

pedrobaeza:
That's clearly not correct. You can't use a computed method for doing things outside the scope. In any case, you can use an ondelete="cascade", or dedicated migration code.

@angelgarciadelachica please remove all that code from the compute.

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please fw-port it to v16.

PR #477, opened 2023-11-22
[14.0][IMP] sale_commission_agent_restrict: add user as demo data

HekkiMelody:


francesco-ooops (approved):
Functional ok!

HekkiMelody:
Merging as it's only demo data, should be fairly safe.

/ocabot merge nobump

PR #480, opened 2023-11-24
[IMP][14.0]sale_commission:Add partial commission settlement

odooNextev:
This PR adds the "Payment Amount Based"" option in "Commission type > Invoice status" field. 

"Payment Based" allows commissions to be added to agent settlement only when the invoice is marked as paid (completely paid); 

"Payment Amount Based" allows commissions to be added to agent settlement proportionally to the payment received, eg: 

Invoice 1 amount 1000 > agent commission on invoice lines = 20% (200) > Invoice payment received 500 > agent commission settled = 100


sale_commission/models/sale_commission_mixin.py
```diff
@@ -124,6 +116,23 @@ class SaleCommissionLineMixin(models.AbstractModel):
...
+                        and x.agent_id == record.agent_id
+                        and x.commission_id.invoice_state != "payment_amount"
+                    )
+                )
+                > 1
+            ):
+                raise exceptions.ValidationError(
+                    _("You can only add one time each agent."),
+                )
+
```
HekkiMelody: **issue**: Maybe I'm mistaken, but I don't think this is correct. `self` only contains the records that have had the fields in `api.constrains` modified. Does this work if you add one object, save, and then add another object in a different operation?

**suggestion**: If I'm indeed correct, then you need to use `.search` instead of just filtering `self`.

sale_commission/tests/test_sale_commission.py
```diff
@@ -391,16 +400,16 @@ def test_commission_propagation(self):
...
+        # # HACK: Remove constraints as in test mode it raises, but not on regular UI
+        # # TODO: Check why this is happening only in tests
+        # self.env.cr.execute(
+        #     "ALTER TABLE sale_order_line_agent DROP CONSTRAINT "
+        #     "sale_order_line_agent_unique_agent"
+        # )
+        # self.env.cr.execute(
+        #     "ALTER TABLE account_invoice_line_agent DROP CONSTRAINT "
+        #     "account_invoice_line_agent_unique_agent"
+        # )
```
HekkiMelody: **chore**: you can probably remove all these lines

HekkiMelody (changes requested):
Partial code review

sale_commission/readme/CONFIGURE.rst
```diff
@@ -21,6 +21,8 @@ For adding commissions:
 
    * **Invoice Based**: Commissions are settled when the invoice is issued.
    * **Payment Based**: Commissions are settled when the invoice is paid.
+   * **Based Payment amount** : Commission are settled when at least one 
+     payment is registered. It allows to settle partial commissions.
```
francesco-ooops: * **Payment Amount Based** : Commissions are settled even when invoice is not fully paid. It allows to settle partial commissions, proportionally to payment received.

sale_commission/models/account_move.py
```diff
@@ -132,6 +132,22 @@ def _copy_data_extend_business_fields(self, values):
...
+            for agent_line in agent_lines:
+                agent_id = agent_line[2]["agent_id"]
+                if agent_id not in agent_id_list:
+                    agent_id_list.append(agent_id)
+                    agent_line[2]["is_partial_settled"] = False
+                    agent_list.append(agent_line)
+            if agent_list:
+                move["agent_ids"] = agent_list
+        return res
+
```
HekkiMelody: **chore**: Please add a comment about what you're doing and why, since this bit of code is quite intricate. Also, we can probably do some code cleanup here, and pick better variable names to increase readability.

**suggestions**:

```suggestion
    def copy_data(self, default=None):
        """
        In case of partial settlements, there may be multiple lines for the same agent
        when copying, preserve only one line per agent.
        """
        res = super().copy_data()
        for move in res:
            old_agent_ids = move.get("agent_ids")
            processed_agents = []
            new_agent_ids = []
            for agent_line in old_agent_ids:
                agent_id = agent_line[2]["agent_id"]
                if agent_id not in processed_agents:
                    processed_agents.append(agent_id)
                    agent_line[2]["is_partial_settled"] = False
                    new_agent_ids.append(agent_line)
            if new_agent_ids:
                move["agent_ids"] = new_agent_ids
        return res

```

sale_commission/models/sale_commission_mixin.py
```diff
@@ -124,6 +116,23 @@ class SaleCommissionLineMixin(models.AbstractModel):
...
+                    [
+                        ("object_id", "=", record.object_id.id),
+                        ("agent_id", "=", record.agent_id.id),
+                        ("commission_id.invoice_state", "!=", "payment_amount"),
+                    ]
+                )
+                > 1
+            ):
+                raise exceptions.ValidationError(
+                    _("You can only add one time each agent."),
```
HekkiMelody: **Suggestion(non-blocking)**:

```suggestion
                    _("Each agent can only be present once - unless the commission type is 'Payment Based'"),
```

sale_commission/models/account_move.py
```diff
@@ -234,3 +251,57 @@ def _skip_future_payments(self, date_payment_to):
             if any(date_payment_to < date for date in payments_dates):
                 return True
         return False
+
+    def _partial_commissions(self):
```
HekkiMelody, line 259: **todo**: Please add a docstring explaining the purpose of this method. I find it hard to follow personally

sale_commission/models/account_move.py
```diff
@@ -234,3 +251,57 @@ def _skip_future_payments(self, date_payment_to):
...
+                )
+                if line.id not in lines_to_update:
+                    lines_to_update[line.id] = {
+                        "amount": line_total_amount - partial_commission
+                    }
+                elif partial_commission >= lines_to_update[line.id]["amount"]:
+                    lines_to_update[line.id] = {
+                        "amount": lines_to_update[line.id]["amount"],
+                        "is_partial_settled": True,
+                    }
```
HekkiMelody: **question**: Here you want to update the existing dictionary, is that correct?

if so, wouldn't it be better to simply write "is_partial_settled"?

e.g. `lines_to_update[line.id].update({"is_partial_settled": True})` or `lines_to_update[line.id]["is_partial_settled"] = True)`

sale_commission/models/account_move.py
```diff
@@ -234,3 +251,57 @@ def _skip_future_payments(self, date_payment_to):
...
+                    break
+                partial_commission_to_settle = line.copy(
+                    {"object_id": line.object_id.id}
+                )
+                partial_commission_to_settle.update(
+                    {"amount": partial_commission, "is_partial_settled": True}
+                )
```
HekkiMelody: **question**: can this be done in a single step?

```suggestion
                partial_commission_to_settle = line.copy(
                    {"object_id": line.object_id.id, 
                    "amount": partial_commission, 
                    "is_partial_settled": True}
                )
```

sale_commission/tests/test_sale_commission.py
```diff
@@ -692,3 +693,27 @@ def test_payment_date_settlement(self):
...
+                (
+                    "agent_id",
+                    "=",
+                    self.agent_monthly.id,
+                ),
+                ("state", "=", "settled"),
+            ]
+        )
+        self.assertEqual(1, len(settlements))
+        self.assertEqual(1, len(settlements.line_ids))
```
HekkiMelody: Could you add a check that there are now two agent lines in the invoice, and that one has `is_partial_settled == True` and the other `False`

Also, if you can do `partial_amount = invoice.amount_total / 4` or `/ 10` you could check the amounts as well (If you split in half, the amounts would be the same)

HekkiMelody (changes requested):
Partial code review.

As always, also squash your commits and add a message following OCA guidelines. https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

HekkiMelody (changes requested):
Code review, LGTM.

Fix the commits and I'll approve

francesco-ooops (approved):
Functional ok!

HekkiMelody (approved):

francesco-ooops:
@pedrobaeza feedback on this? thanks!

pedrobaeza (changes requested):
My take on this is that it's a very convoluted case to be included in the base module, as it adds a complexity to maintain it, and migrating the module won't assure that those without the use case will handle it correctly.

You can add a new module for this case, adding the proper hooks. Comments on the implementation:

- Using `invoice_state` for adding a new case is not the correct approach, as that field is for when, not for which amount, and you are mixing both. It only means that any reference on one part of the code to `invoice_state == "paid"` will not be valid and you will need to add specific code, breaking extensibility. Create a new field for your purpose with 2 values (full amount or paid amount), having the first one as the default behavior.
- Removing the unicity constraint is not OK. Apart from changing the current DB scheme, using this approach will denormalize the expected structure, breaking again the extensibility. You will need to create a super-structure or other way to do it.

odooNextev:
@pedrobaeza Thanks! I'm currently work on a new modulo to add this feature.
About your consideration regarding the unicity constraint I cannot undestand how is it possible to not respect that unicity condition. I Tried so many times to generate the unique error from user interface but it looks not possible.

pedrobaeza:
I mean that the unique constraint shouldn't be changed to Python constraint, and should be preserved. I suppose you have removed it because now the criteria for unique has changed. That's what I'm saying that you shouldn't vary, and design the datamodel other way.

odooNextev:
Thanks Pedro, I understand what you mean. But I don't undestand what's the reason to have a unique constraints between agent_id and account_invoice_line, why is it not possible to have multiple agent line connected to same line?
And how can I reproduce it from user interface?

pedrobaeza:
In this case is more a question of the initial design choices and stability policy. Whatever reason for that, the problem is that if you change this schema in the middle of the release, some extensions (or the module itself) may count with having only one line per agent, so failing when with your new module, you change that rules.

Let's say "you can't change the rules in the middle of the game".

Was that one a good design decision? Maybe not, maybe yes. It was thought for not requiring more data to determine a line, as putting other criteria (like commission type) would complicate the identification and not having to sum commission amounts to get the total. If you need specific computations, you put a formula for that. This can be questioned and changed on 17.0, but not for the current existing versions.

For your case, I recommend to put a children table for doing such split. In fact, that design may be more normalized according usual DB normalization rules.

odooNextev:
superseded #486

PR #481, opened 2023-11-24
[14.0][FIX] sale_commission_product_criteria_domain: multicompany item agent

HekkiMelody:
This commit fixes a multicompany error that happened when the partner's allowed companies did not match with the agent's.

In such case (especially common when using the module partner_multi_company) the user could see the partner but could not see the agent assigned to the partner; therefore the user also shouldn't see the commission_item_agent_ids field that involve either a partner or an agent the user shouldn't see.


francesco-ooops (approved):
Functional ok!

@ilyasProgrammer what do you say?

ilyasProgrammer (approved):
LGTM

PR #483, opened 2023-11-28
[16.0][FIX] account_commission: mandatory fields cannot be set to False.

angelgarciadelachica:
Same FIX that in v15.0 https://github.com/OCA/commission/pull/476


pedrobaeza (approved):
Fast-tracking this:

/ocabot merge patch

PR #488, opened 2023-12-13
[IMP]Sale_commission_partial_settlement: separate method for skipping…

odooNextev:
… future payments

### DIFF
--- a/sale_commission_partial_settlement/models/account_partial_reconcile.py
+++ b/sale_commission_partial_settlement/models/account_partial_reconcile.py
@@ -5,3 +5,8 @@ class AccountPartialReconcile(models.Model):
     _inherit = "account.partial.reconcile"
 
     partial_commission_settled = fields.Boolean()
+
+    def _skip_future_partial_payments(self, date_payment_to, counterpart_line_date):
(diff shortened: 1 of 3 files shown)


francesco-ooops:
@pedrobaeza opinions on this?

sale_commission_partial_settlement/models/account_partial_reconcile.py
```diff
@@ -5,3 +5,8 @@ class AccountPartialReconcile(models.Model):
     _inherit = "account.partial.reconcile"
 
     partial_commission_settled = fields.Boolean()
+
+    def _skip_future_partial_payments(self, date_payment_to, counterpart_line_date):
```
pedrobaeza, line 9: There's not too much sense to put here this method, as `self` is not used at all.

HekkiMelody (commented):
Code review ok, aside from what has already been pointed out.

Another nitpick is that the commit message should be shorter for the first commit; in the second commit it should mention the module name. See: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

PR #490, opened 2023-12-18
[14.0] [IMP] sale_commission: Clarify selection fields

renda-dev:
With this PR we aim to make the selection fields as clear as possible, adding help text and rewording the `amount_base_type` selections.


francesco-ooops (approved):
LGTM!

@pedrobaeza Gross/Net definitions (especially when terms are translated) sound related to taxed/untaxed values. I think this is a useful clarification to port to upper version

pedrobaeza (approved):
OK, agreed.

/ocabot merge minor

PR #493, opened 2024-01-04
[15.0][IMP] account_commission: Improve settlement report

pedrobaeza:
Switch invoice line column to Invoice name + Customer + Description + Amount for better comprehension.

@Tecnativa TT46914

### DIFF
--- a/account_commission/i18n/account_commission.pot
+++ b/account_commission/i18n/account_commission.pot
@@ -59,6 +59,7 @@ msgstr ""
 
 #. module: account_commission
 #: model:ir.model.fields,field_description:account_commission.field_invoice_commission_analysis_report__amount
+#: model_terms:ir.ui.view,arch_db:account_commission.report_settlement_document
 msgid "Amount"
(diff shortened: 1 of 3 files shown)


chienandalu (approved):

jguenat (commented):
price_subtotal might not be relevant depending on how the commission is calculated but I agree it is useful for gross amount based commission, which is the generic case.

Could add a condition on the price_subtotal column but hard to do because there can be multiple commissions types.

what do you think?

pedrobaeza:
It wouldn't be performant-wise to explore all the commission lines to see if there's any commission of such type. But the same problem is for quantity that you previously added: for my customers, it's not useful, but as it takes very few, let's add both and it can be ignored if not needed.

jguenat (approved):
Okay let's merge like that. Specific needs can be addressed in client specific custo

pedrobaeza:
Thanks for your comprehension.

/ocabot merge minor

PR #495, opened 2024-01-10
[14.0][FIX] sale_commission_advance: fix partner in method call

HekkiMelody:
Small fix to pass the partner to the method


HekkiMelody:
My bad, I got confused while fixing something else.

PR #502, opened 2024-01-22
[fix] [sale_commision] fix search  field partner_agent_ids on multi company

matteoopenf:
I have the following problem.
I have an installation of odoo with 9 companys, I have three agents. The agent is agent for more contact in every company.
So if I try from UI to use that searc, I obtain the following error:

image
https://github.com/OCA/commission/blob/6c7c904711397dc93d06801d93533839194904a7/sale_commission/models/sale_order.py#L25

-Create one o more agent
- Create 2 or more company
- create some contact in every company ad assign the same agent
- Now create an order in every company to a contact have the sames agents
- Now try to search in base on field partner_agent_ids and have the problem on the screen

with this fix after recover all record  sale.order.line.agent respect the search, I filter the record in base of the company the user is logged

### DIFF
--- a/sale_commission/models/sale_order.py
+++ b/sale_commission/models/sale_order.py
@@ -34,8 +34,16 @@ class SaleOrder(models.Model):
 
     @api.model
     def _search_agents(self, operator, value):
+
+        # Note: we need add in the domain company_id to allow retrieve only
+        # the records of the company the user is logged otherwise we have
+        # problem with this rule: https://shorturl.at/pwBIJ
+
         sol_agents = self.env["sale.order.line.agent"].search(
-            [("agent_id", operator, value)]
+            [
+                ("agent_id", operator, value),
+                ("object_id.company_id", "=", self.env.company.id),
+            ]
         )
         return [("id", "in", sol_agents.mapped("object_id.order_id").ids)]
 


patrickt-oforce (approved):

amosdisabatino (approved):

pedrobaeza (changes requested):
- Please don't add empty lines inside a method.
- The search should be for all the allowed companies, not the main one.

matteoopenf:
> * Please don't add empty lines inside a method.
> * The search should be for all the allowed companies, not the main one.

Hi @pedrobaeza , 

ok for the lines empty.
About "The search should be for all the allowed companies, not the main one." if have not that fix I receive  a securty error asI show in the screen

pedrobaeza:
Well, the error should be in other part then, as allowed companies should be supported.

matteoopenf:
> Well, the error should be in other part then, as allowed companies should be supported.

I did a lot of debug to avoid the problem on the screen, but the error is caused because we try to  access to record of other company.

sale_commission/models/sale_order.py
```diff
@@ -34,8 +34,16 @@ def _compute_agents(self):
...
+
+        # Note: we need add in the domain company_id to allow retrieve only
+        # the records of the company the user is logged otherwise we have
+        # problem with this rule: https://shorturl.at/pwBIJ
+
         sol_agents = self.env["sale.order.line.agent"].search(
-            [("agent_id", operator, value)]
+            [
+                ("agent_id", operator, value),
+                ("object_id.company_id", "=", self.env.company.id),
```
matteoopenf, line 45: @pedrobaeza you mean  here should be ("object_id.company_id",   'in',  self.env.companies.ids) ?

pedrobaeza: Yes

PR #507, opened 2024-03-17
[16.0][IMP]Place footer outside the table

flotho:
so that it won't be repeated at each page

### DIFF
--- a/account_commission/views/report_settlement_templates.xml
+++ b/account_commission/views/report_settlement_templates.xml
@@ -38,11 +38,5 @@
                 </td>
             </t>
         </xpath>
-        <xpath expr="//table/tfoot/tr/td" position="before">
-            <t t-if="o.settlement_type == 'sale_invoice'">
-                <td />
-                <td />
-            </t>
-        </xpath>
     </template>
 </odoo>
--- a/commission/reports/report_settlement_templates.xml
+++ b/commission/reports/report_settlement_templates.xml
@@ -51,21 +51,31 @@
                                 </td>
                             </tr>
                         </tbody>
-                        <tfoot>
-                            <tr>
-                                <td />
-                                <td class="text-end">
-                                    <strong>Total</strong>
-                                </td>
-                                <td class="text-end">
-                                    <span
-                                        t-field="o.total"
-                                        t-options="{'widget': 'monetary', 'display_currency': o.currency_id}"
-                                    />
-                                </td>
-                            </tr>
-                        </tfoot>
                     </table>
+                    <div class="mb-4">
+                        <div id="total" class="row">
+                            <div
+                                t-attf-class="#{'col-6' if report_type != 'html' else 'col-sm-7 col-md-6'} ms-auto"
+                            >
+                                <table
+                                    class="table table-sm table-borderless"
+                                    style="page-break-inside: avoid;"
+                                >
+                                    <tr class="border-black o_total">
+                                        <td>
+                                            <strong>Total</strong>
+                                        </td>
+                                        <td class="text-end">
+                                            <span
+                                                t-field="o.total"
+                                                t-options="{'widget': 'monetary', 'display_currency': o.currency_id}"
+                                            />
+                                        </td>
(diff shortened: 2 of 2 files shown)


flotho:
Hi @pedrobaeza ,

Small proposal here to prevent footer to be displayed on each page.
Fairly common structure like in sale.order, account.invoice reporting.

Thanks for the time you'll spent on this review

commission/reports/report_settlement_templates.xml
```diff
@@ -9,7 +9,7 @@
                         <address
                             class="mb-0"
                             t-field="o.agent_id"
-                            t-options='{"widget": "contact", "fields": ["address", "name"], "no_marker": True}'
+                            t-options="{'widget': 'contact', 'fields': ['address', 'name'], 'no_marker': True}"
```
pedrobaeza: Why changing this?

account_commission/views/report_settlement_templates.xml
```diff
@@ -38,11 +38,5 @@
                 </td>
             </t>
         </xpath>
-        <xpath expr="//table/tfoot/tr/td" position="before">
```
pedrobaeza, line 41: Why removing this?

commission/reports/report_settlement_templates.xml
flotho: you're right, my VScode change this pattern inappropriately

account_commission/views/report_settlement_templates.xml
flotho: this inheritance add extra cells to be able to align the total on right whatever the content.
Yet AFAICS, using tfoot repeat the footer on each page so using tfoot looks inefficient to me.
And when using a dedicated table for the total summary, no need to use extra cell to be aligned.

pedrobaeza: But is it used in any dependent module? If so, it should be changed there as well.

pedrobaeza:
Please don't add extra commits, but squash/amend them into one.

account_commission/views/report_settlement_templates.xml
flotho: actually, this inheritance in the account module is done because of the structure of the main view.
Depending of the module, it adds extra columns. That's why account_commission was adding a new column.
Now, with the simple syntax outside the table, there is no more problems for displaying properly the total, whatever the module adding new columns.

flotho:
> Please don't add extra commits, but squash/amend them into one.

done !

commission/reports/report_settlement_templates.xml
```diff
@@ -51,21 +51,31 @@
...
-                                <td class="text-end">
-                                    <span
-                                        t-field="o.total"
-                                        t-options="{'widget': 'monetary', 'display_currency': o.currency_id}"
-                                    />
-                                </td>
-                            </tr>
-                        </tfoot>
                     </table>
+                    <div class="clearfix mb-4">
```
pedrobaeza: Why are you putting `clearfix`? That kind of dirty patches amending the whole thing should be avoided.

commission/reports/report_settlement_templates.xml
```diff
@@ -51,21 +51,31 @@
...
+                                    class="table table-sm table-borderless"
+                                    style="page-break-inside: avoid;"
+                                >
+                                    <tr class="border-black o_total">
+                                        <td>
+                                            <strong>Total</strong>
+                                        </td>
+                                        <td class="text-end">
+                                            <span
+                                                t-field="o.total"
```
pedrobaeza, line 70: Isn't `total` already a monetary field for avoiding to put the `t-options`? And if not, better to convert it to monetary.

pedrobaeza (commented):
Please name the commit properly:

`[IMP] *commission: Report footer outside the table`

flotho:
done @pedrobaeza

pedrobaeza (approved):

gaelTorrecillas (approved):

PR #511, opened 2024-04-02
[15.0][FIX] account_commission: use Form() in tests

chienandalu:
Not using Form() in tests can lead to some tricky integration tests issues. In this case, just overriding _recompute_payment_terms_lines with a new context in a third module wasn't making these true.

cc @Tecnativa TT47134

please review @pedrobaeza @victoralmau @Tisho99


pedrobaeza (approved):

Tisho99 (approved):
LGTM. Thanks for the fix!

victoralmau (approved):

PR #512, opened 2024-04-02
[15.0][FIX] sale_commission: ensure currency

chienandalu:
We should ensure that the partner's currency is the one in the company to avoid test integration issues

cc @Tecnativa TT38303

please review @pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #513, opened 2024-04-03
[16.0][FIX] sale_commission: ensure currency

chienandalu:
fw of https://github.com/OCA/commission/pull/512

We should ensure that the partner's currency is the one in the company to avoid test integration issues

cc @Tecnativa TT38303

please review @pedrobaeza @victoralmau


pedrobaeza (approved):
Fast-tracking this

/ocabot merge patch

PR #466, opened 2023-10-19
[16.0][MIG] sale_commission_product_criteria_discount

SirAionTech:
Migration from `14.0` (https://github.com/OCA/commission/tree/0cb07b1fe6e6a07db5f5a9a823d5e593b8be26c2/sale_commission_product_criteria_discount).

PR #467, opened 2023-10-19
[16.0][MIG] sale_commission_product_criteria_domain

SirAionTech:
Migration from `14.0` (https://github.com/OCA/commission/tree/0cb07b1fe6e6a07db5f5a9a823d5e593b8be26c2/sale_commission_product_criteria_domain).

I have included a couple of **FIX**es that I have noticed while looking at the module code, I'll backport them when this is merged.

I had to change the demo data because of conflicts with other modules, I added more details in the **[MIG]** commit description.

Depends on:

- [x] https://github.com/OCA/commission/pull/466
      This one is not a dependency of the module itself, but I have noticed that tests fail if the **_discount** module is missing.
      And it looks like adding this in **test-requirements.txt** is not enough because the **_discount** module is not tested.
      After deeper investigation, what happens is that the agent line of the order line matches the commission_type `sale_commission_product_criteria_discount.demo_commission_rules_item_disc_2`, resulting in the tests failure:
      
     > Traceback (most recent call last):
     File "/path/to/commission/sale_commission_product_criteria_domain/tests/test_sale_commission_product_cri>teria_domain.py", line 258, in test_commission_domain
          self.assertEqual(so.order_line.agent_ids.amount, 15)
       AssertionError: 10.0 != 15

     because the fixed_discount of the matched commission item is 15 but if the module **sale_commission_product_criteria_discount** is not installed, another commission item is matched (`sale_commission_product_criteria.demo_commission_rules_item_1`) that has fixed_discount 10.

     **TL;DR**: test will fail until https://github.com/OCA/commission/pull/466 is merged, test-requirements is no use.

PR #514, opened 2024-04-04
[ADD] sale_commission_product_criteria_triple_discount

SirAionTech:
Depends on:
- [x] https://github.com/OCA/commission/pull/466


SirAionTech:
Tests of **sale_commission_product_criteria** are failing with error
  > 2024-04-04 08:24:47,763 463 ERROR odoo odoo.addons.sale_commission_product_criteria.tests.test_sale_commission_product_criteria: FAIL: TestSaleCommission.test_sale_commission_product_criteria_items
Traceback (most recent call last):
  File "/__w/commission/commission/sale_commission_product_criteria/tests/test_sale_commission_product_criteria.py", line 80, in test_sale_commission_product_criteria_items
    self.assertEqual(so_1.order_line.agent_ids.amount, 10)
AssertionError: 15.0 != 10
  
  because the matched commission item is `sale_commission_product_criteria_discount.demo_commission_rules_item_disc_2` instead of `sale_commission_product_criteria.demo_commission_rules_item_1`, this happens in https://github.com/OCA/commission/blob/35a4363d6178a2b6ee0002556e9b4f1fc9e8abbc/sale_commission_product_criteria/models/sale_commission_line_mixin.py#L62.
It is something similar to what already described in https://github.com/OCA/commission/pull/467.
I don't exactly understand why it is failing (and how it is working in `14.0`), will investigate further when the dependency is merged.

SirAionTech:
Rebased on `16.0`.

PR #518, opened 2024-04-20
add es.po

berisoft-arg:


pedrobaeza:
Please use Weblate platform for this:

https://odoo-community.org/resources/translate

You can upload .po files there as well.

PR #519, opened 2024-04-23
[15.0][IMP] commission: Add new settlement period (All pending records) to settle in only one settlement

sergio-teruel:
cc @Tecnativa TT48574

ping @carlosdauden @victoralmau


commission/wizards/commission_make_settle.py
```diff
@@ -97,7 +103,7 @@ def _prepare_settlement_line_vals(self, settlement, line):
             "settlement_id": settlement.id,
         }
 
-    def _get_agent_lines(self, date_to_agent):
+    def _get_agent_lines(self, agent, date_to_agent):
```
pedrobaeza, line 106: Why do you need to add this argument? This is breaking possible extensions.

sergio-teruel: I do not know... 
All modules that extends this method are changing the signature... So I change the signature in original method

pedrobaeza: Ah, really? Then it can be considered a bug. Can you link one of those overrides?

sergio-teruel: https://github.com/OCA/commission/blob/3a55b78b7f4773771577dc83d12d129a08147be1/account_commission/wizards/commission_make_settle.py#L17

carlosdauden (approved):

commission/wizards/commission_make_settle.py
pedrobaeza: Then it's definitively a bug here, so it's OK to change it indeed, thanks.

pedrobaeza (approved):
The only risk here is that you don't settle on time (accumulating several months), and you get everything in only one settlement.

victoralmau (approved):

pedrobaeza:
Please fw-port it to 16 and 17

/ocabot merge minor

PR #526, opened 2024-05-14
[17.0] [FIX] commission: partner view agent

Tisho99:
In v17, the commission module has an incompatibility with the base_vat module, because the "vat" field of the partner form is moved inside a new div in [its views](https://github.com/odoo/odoo/blob/e0ee100b55c58a8d0968cf5c536a4cdb0d72b667/addons/base_vat/views/res_partner_views.xml#L16), and the "agent" field is placed before that field inside the commission module.

As a result, the "agent" field is displayed without tag in the partner form

image

This PR edits the xpath to always put the field at the end of the vat group

T-5880

### DIFF
--- a/commission/views/res_partner_views.xml
+++ b/commission/views/res_partner_views.xml
@@ -6,9 +6,9 @@
         <field name="inherit_id" ref="base.view_partner_form" />
         <field name="priority" eval="18" />
         <field name="arch" type="xml">
-            <field name="vat" position="after">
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

Tisho99:
@ValentinVinagre @HaraldPanten 

@luis-ron showed me this error last week

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):

PR #527, opened 2024-05-15
[14.0] [FIX] sale_commission_geo_assign: reset item_agent_ids on replace

renda-dev:
Without resetting this lines, the commission_item_group doesn't get updated properly, so to get the right lines you would have to manually delete them and then run the wizard again.

This PR fixes the explained issue

edit: <del>I'm also adding me and my team as maintainers of this module</del> (added maintainers in https://github.com/OCA/commission/pull/529)


francesco-ooops (approved):
Functional ok!

The feature was not working as expected. @eLBati would you mind merging this? We're added as maintainers so we're independent if there's anything else to fix. Thanks!

HekkiMelody (approved):
Code review, LGTM.

Not the most elegant solution but...

PR #535, opened 2024-06-14
[IMP] *commission: update settlement, tables and models names

CLaurelB:
commission module:
Add migration scripts to update the settlement and commision references in the account.move.line and commission.settlement.line models. Also, add a migration script to rename commission tables and models to match the new ones defined in the module.

account_commission module:
Add a migration script to update the settlement line data to match the new field type defined in the module.

### DIFF
--- a/account_commission/migrations/17.0.1.0.0/post-migration.py
+++ b/account_commission/migrations/17.0.1.0.0/post-migration.py
@@ -1,0 +1,20 @@
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Convert the former `agent_line` m2m relation in `commission.line.mixin` into
+    the new `settlement_line_ids` o2m relation."""
+    openupgrade.logged_query(
+        env.cr,
+        """
+            UPDATE commission_settlement_line
+            SET invoice_agent_line_id = sal_rel.agent_line_id
+            FROM (
+                SELECT DISTINCT ON (agent_line_id) agent_line_id, settlement_id
+                FROM settlement_agent_line_rel
+                ORDER BY agent_line_id
+            ) AS sal_rel
+            WHERE id = sal_rel.settlement_id
+        """,
+    )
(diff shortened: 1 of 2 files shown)


CLaurelB:
@luisg123v could you review, please?

pedrobaeza (changes requested):
These changes were already done where they belong, in 15.0:

https://github.com/OCA/commission/blob/15.0/commission/migrations/15.0.2.0.0/pre-migration.py

If you are doing an enterprise migration directly from 14.0-, please don't publish things here.

It's also good to respect copyright from others' work.

luisg123v:
@pedrobaeza,

> If you are doing an enterprise migration directly from 14.0-, please don't publish things here.

Could you elaborate on this, please? I'd like to understand the correct way to do this in the OCA and why.

> It's also good to respect copyright from others' work.

I don't think @CLaurelB aims to take credit over someone else's work.

Regards,

pedrobaeza:
The migration scripts are already done in its corresponding version where the DB schema changed. You can't put the same migration scripts here in this version because you are jumping using the enterprise migration blackbox from 14.0- to 17.0.

These migration scripts are not idempotent, so that means that this, executed several times, can break things. Users coming from 16.0 or from previous one but using OpenUpgrade will have incorrect migrations due to them.

You have to gather all the OCA modules migration scripts (and maybe adapt them) on your own, but you can't put them on OCA, as there can be as many scripts as combinations of the source and target version (from 13 to 17, from 14 to 17, from 13 to 16, from 14 to 16, etc, etc).

luisg123v:
@pedrobaeza thanks a lot for the explanation, that's just what I was suspecting.

ANd sorry, this was asked by myself, @CLaurelB was just following my advice on this.

I'll keep it in mind for future migrations.

Regards,

pedrobaeza:
At the end, migration through enterprise service with a lot of OCA modules doesn't worth too much due to this work, but it's your choice.

PR #544, opened 2024-07-17
[14.0][FIX]sale_commission: settlement report fixes

PicchiSeba:
EDIT:

This PR contains the following changes to the settlement report

- fixes to the spacing of certain elements
~~removed the field "Invoice" because it is always empty~~


sale_commission/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    show_partner_settlement_report = fields.Boolean(
+        "Show partner details in settlements report",
+        default=False,
+    )
+    settlement_skip_ero_amount_lines = fields.Boolean(
```
francesco-ooops: typo (also elsewhere)

francesco-ooops (approved):
Functional ok!

in this PR:

- fixed settlement report ("Invoice" field was empty but was not relevant as the same info are present in "Invoice line" field)
- Added option to display partner in settlement lines report
- Added option to not display settlement lines with zero amount (useful with _product_criteria modules which generate many such lines)

sale_commission/readme/CONFIGURE.rst
```diff
@@ -48,3 +48,17 @@ For adding new agents:
...
+#. Go to *Settings > Sales > Quotations & Orders*:
+
+  * enable he setting "Show partner details in settlements report"
+
+
+To hide lines with amount equals to 0:
+
+#. Go to *Settings > Sales > Quotations & Orders*:
+
+  * enable he setting "Skip lines in settlements with zero amount"
```
renda-dev: **typo:**

```suggestion
  * enable the setting "Show partner details in settlements report"


To hide lines with amount equals to 0:

#. Go to *Settings > Sales > Quotations & Orders*:

  * enable the setting "Skip lines in settlements with zero amount"
```

renda-dev (approved):
Code review, LGTM

francesco-ooops:
@pedrobaeza what do you say? :)

pedrobaeza (changes requested):
For me this is overcomplicating the module introducing computed fields that slows down all the usual workflows. This should be IMO in a separated module for those wanting that specific behavior.

Apart from that, computed one2many are not officially supported, and can give some glitches. They should be m2m.

francesco-ooops:
@pedrobaeza shall we at least leave the commit where we fix report?

pedrobaeza:
OK.

PicchiSeba:
Ok @pedrobaeza.

I removed all the unnecessary changes from this PR (I will eventually create a new module in a new PR). I only left the fixes regarding spacing in the report

francesco-ooops:
@PicchiSeba please reinstate the fix with the removal of column "invoice" from report

PicchiSeba:
I also updated the PR's title and description

pedrobaeza:
Now reading the restricted fix, this is not really true. The `invoice_id` field is filled with the vendor bill that is done for outsiders. Your DB will be empty if all the commission agents are internal, but we shouldn't remove it.

PicchiSeba:
That's exactly our case. Good to know. I'm removing the offending commit

pedrobaeza (approved):
/ocabot merge minor

francesco-ooops:
> Now reading the restricted fix, this is not really true. The `invoice_id` field is filled with the vendor bill that is done for outsiders. Your DB will be empty if all the commission agents are internal, but we shouldn't remove it.

@pedrobaeza I think our use case is a settlement report where agent vendor bill hasn't been created yet, so that condition could exclude the field from the report. We'll think of such change

pedrobaeza:
Maybe you can detect if there's any invoice in whole report, and it not, hide the column.

PR #546, opened 2024-07-17
[16.0][MIG] sale_commission_product: Migration to 16.0

odooNextev:
I don't know exactly if it can be defined as a migration itself, but I started from `sale_commission_product` module for this project, I kept commits history and mostly the goal of my module is the same as initial version `8.0`.

In fact this module allows users to set dedicated agents on `product.template` (unlike `8.0` version which added the agent field on `product.product`) and they will find in `sale.order.line` both agents on SO partner and agents on SOL product.

Talking about computation, I simplified what was done in `8.0` version and I hope it can be enough.
In particular I only extended `_prepare_agents_vals_partner` method to add product agents to partner agents array returned from original `sale_commission` method.
To trigger this method on product changing, I had to add `@api.depends("product_id")` to `_compute_agent_ids` of `sale.order.line` and `account.move.line`.
By the way, I followed what was done in `8.0` version and I added account section even if `_compute_agent_ids` method of `account.move.line` it's inside `account_commission` and not in `sale_commission`.
Alternatively we could change the module name to include "account" or create a separated module like `account_commission_product`, but I don't know if it's worth it.
Anyway I added `account_commission` to depends.


francesco-ooops:
@odooNextev can you add docs files?

odooNextev:
> @odooNextev can you add docs files?

Is it enough?

francesco-ooops:
Looks good thanks!

francesco-ooops (approved):
Functional ok

PR #552, opened 2024-08-08
[15.0][FIX] *commission: Remove related from commission_free

pedrobaeza:
Steps to reproduce:

- Create a product commission free.
- Do some invoices with that product.
- There's no commission on the invoices.
- Remove the "Commission free" check.

Result: all the past invoices are recomputed and now a commission appears. That's not acceptable, as past things shouldn't be changed if you change the definition from now on, and more if you already did the settlements.

This happens because the field in the invoice line is a related one, so changing the source, recomputes all the linked records.

Let's change it to a computed writable, being recomputed only when the product changes, not if we change the .

@Tecnativa TT50445


CarlosRoca13 (approved):
LGTM :+1:

chienandalu (approved):

PR #553, opened 2024-08-08
[16.0][IMP] account_commission: Index on invoice_agent_line_id

pedrobaeza:
A little improvement for indexing a column that it's used as FK. That way, deletions of agent lines in invoices are faster.

@Tecnativa TT50445


CarlosRoca13 (approved):

victoralmau (approved):

chienandalu (approved):

pedrobaeza:
It turns out that a similar change was already implemented in 16.0, so only the index is needed.

/ocabot merge patch

PR #558, opened 2024-08-22
[15.0][FIX] sale_commission_salesman: partner fetch for agents creation on move lines

CLaurelB:
Fetch the partner by move line record instead of fetching all partners for all invoice users related to the moves. The previous method could fetch more than one partner, including those not related to the move, when multiple lines for different moves are being processed, causing an error when trying to obtain the values of the `agent` and `salesman_as_agent` fields.

Back-port of 9a2951db


pedrobaeza (approved):
Thank you!

Fast-tracking it as already merged in 16.0

/ocabot merge patch

PR #565, opened 2024-10-10
[15.0][IMP] commission: add the commission_ids field to the settlement and include it in the views

AungKoKoLin1997:
This PR improves the user experience by making it easier to understand which commission types are assigned to each settlement record, instead of having to check each settlement line individually.

@qrtl QT4840

### DIFF
--- a/commission/models/commission_settlement.py
+++ b/commission/models/commission_settlement.py
@@ -60,6 +60,14 @@ class CommissionSettlement(models.Model):
         default=lambda self: self._default_company_id(),
         required=True,
     )
+    commission_ids = fields.Many2many(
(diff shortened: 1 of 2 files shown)


yostashiro (approved):
Code review.

pedrobaeza (changes requested):
Adding a new computed stored field will provoke a collapse updating the module. And anyway, a many2many is not seen very well in a tree view. I prefer this to be done on a custom module.

AungKoKoLin1997:
Closing this PR and will create custom module for this.
@pedrobaeza  Thanks for your suggestion.

PR #568, opened 2024-10-15
[14.0][FIX]sale_commission_product_criteria: ignore error on create

PicchiSeba:
Clicking the "Create" button for commision types, the exception at the end of `check_type_change_allowed_moves` is  raised, without even landing in the commission type form.
In this PR we make sure the commission exists in the first place


francesco-ooops (approved):
Functional ok!

PR #570, opened 2024-10-18
[14.0][IMP] sale_commission: print partial and document total in report

HekkiMelody:

PR #572, opened 2024-10-23
[15.0][IMP] sale_commission: print partial and document total in report

toita86:
FWP of https://github.com/OCA/commission/pull/570 the implementation is a little bit different due to other modules dependance on the position of certain elements in the report.

### DIFF
--- a/commission/reports/report_settlement_templates.xml
+++ b/commission/reports/report_settlement_templates.xml
@@ -17,6 +17,15 @@
             <div class="mt-5">
                 <div class="page">
                     <h2>Settlement</h2>
+                    <t t-if="show_doc_total">
+                        <div class="text-right">
+                            <strong>Global commission amount:</strong>
+                            <span
+                                t-esc="sum(docs.mapped('total'))"
+                                t-options="{'widget': 'monetary', 'display_currency': docs[0].currency_id}"
+                            />
+                        </div>
+                    </t>
                     <div id="informations" class="row mt-4 mb-4">
                         <div class="col-auto col-3 mw-100 mb-2" t-if="o.date_from">
                             <strong>From:</strong>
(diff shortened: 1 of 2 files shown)


toita86:
Hi @aleuffre, while FW you feature to adapt it to v15.0 I needed to change the position of the `div`  that contains the total.

but now looks like this:
image

Do you have any suggestion that could make it more similar to you implementation?

HekkiMelody:
> Hi @aleuffre, while FW you feature to adapt it to v15.0 I needed to change the position of the `div` that contains the total.
> 
> but now looks like this:
> Do you have any suggestion that could make it more similar to you implementation?

I took heavy inspiration from a similar section in the Sales Order report. As you can see, what I did was put the total outside the `table` and `tfoot` element, because the `tfoot` like the `thead` is printed again on every page, which is something we wanted to avoid.

So yes, I would try to put the total in a div, outside the `table` itself

toita86:
> So yes, I would try to put the total in a div, outside the table itself

At first i didn't wanted to move it outside because the module `account_commission` used the `tfoot` on to add some spaces https://github.com/OCA/commission/blob/40491a3c2b1d86d6a3ab28a7e0997e3f8f907216/account_commission/views/report_settlement_templates.xml#L41

by removing the xpath element added in `account_commission` and moving the div outside everything looks fine in IMO.
Now this is the output
image

commission/reports/report_settlement_templates.xml
```diff
@@ -17,6 +17,15 @@
...
                     <h2>Settlement</h2>
+                    <t t-if="show_doc_total">
+                        <div class="text-right">
+                            <strong>Global commission amount:</strong>
+                            <span
+                                t-esc="sum(docs.mapped('total'))"
+                                t-options="{'widget': 'monetary', 'display_currency': docs[0].currency_id}"
+                            />
+                        </div>
+                    </t>
```
HekkiMelody, line 28: **issue**: Since this is the total for all pages being printed, I don't think it should be printed inside the `foreach` (line 95), as you only want this printed once, not once for every commission.

toita86: Thank you for your observation. While it's true that the total is printed inside the for each loop, it actually maintains the same behavior as in the 14.0.

HekkiMelody (approved):
Code review, LGTM

francesco-ooops (approved):
LGTM

@pedrobaeza merge?

pedrobaeza (approved):
/ocabot merge minor

PR #581, opened 2024-12-05
[IMP] account_commission: Settle based on payment date

SirAionTech:
Add a new "Invoice Status" to settle based on the payment date.
For instance: if the invoice is in January but its payment is in February, the invoice will be settled in February's settlement.

### DIFF
--- a/account_commission/README.rst
+++ b/account_commission/README.rst
@@ -48,8 +48,10 @@ For selecting invoice status in commissions:
 
    * **Invoice Based**: Commissions are settled when the invoice is issued.
    * **Payment Based**: Commissions are settled when the invoice is paid or refunded.
-     Note that when refunding an invoice, the corresponding reversed commission will
-     be settled as well, resulting in a 0 net commission between both operations.
+   * **Payment Date Based**: Commissions are settled based on the payment date.
+
+   Note that when refunding an invoice, the corresponding reversed commission will
+   be settled as well, resulting in a 0 net commission between both operations.
 
 Usage
 =====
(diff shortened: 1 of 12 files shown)


francesco-ooops:
@SirAionTech could the same goal be achieved through https://github.com/OCA/commission/pull/441 ?

SirAionTech:
> @SirAionTech could the same goal be achieved through #441 ?

Hi!
Unfortunately I didn't see that, and yes it looks like it is very similar to what happens in this PR with a few differences:

- Here the new logic is decided in the commission type, while in #441 it's set during each settlement.
- Here you don't have to input the new date "Payment date up to" that is used #441, (I think I'd be confused about the interaction of the two date fields) because the existing date field is used to find the payments for the commissions that require it.

I can't say which approach is best, I guess we'll let reviewers and maintainers decide, but thanks for letting me know of that implementation

francesco-ooops:
@SirAionTech I'm missing some context about what you're trying to achieve here: 

The only change from option "Payment Based" is that commission is moved to a different settlement? 

Can you add a functional example highlighting differences between "Payment based" and "Payment date based"?

SirAionTech:
> @SirAionTech I'm missing some context about what you're trying to achieve here:
> 
> The only change from option "Payment Based" is that commission is moved to a different settlement?
> 
> Can you add a functional example highlighting differences between "Payment based" and "Payment date based"?

Sure:

1. Create and confirm an invoice in **January**
2. Pay the invoice in **February**

If the commission is "Payment date based" (added in this PR): the invoice will be settled in **February**'s settlement.
If the commission is "Payment based": the invoice will be settled in **January**'s settlement.

PR #588, opened 2024-12-20
[16.0[IMP]account_commission: add payment date limit on settle commissions

matteotognini:
Porting #441

### DIFF
--- a/account_commission/models/account_move.py
+++ b/account_commission/models/account_move.py
@@ -251,7 +251,28 @@ class AccountInvoiceLineAgent(models.Model):
         :return: bool
         """
         self.ensure_one()
+        payment_based_commission = self.commission_id.invoice_state == "paid"
+        if payment_based_commission and self._skip_future_payments():
+            return True
         return (
-            self.commission_id.invoice_state == "paid"
+            payment_based_commission
             and self.invoice_id.payment_state not in ["in_payment", "paid", "reversed"]
         ) or self.invoice_id.state != "posted"
+
+    def _skip_future_payments(self):
+        date_payment_to = self.env.context.get("date_payment_to")
+        if date_payment_to:
+            payments_dates = []
+            (
+                invoice_partials,
+                exchange_diff_moves,
+            ) = self.invoice_id._get_reconciled_invoices_partials()
+            for (
+                _partial,
+                _amount,
+                counterpart_line,
+            ) in invoice_partials:
+                payments_dates.append(counterpart_line.date)
+            if any(date_payment_to < date for date in payments_dates):
+                return True
+        return False
--- a/account_commission/wizards/commission_make_settle.py
+++ b/account_commission/wizards/commission_make_settle.py
@@ -13,6 +13,12 @@ class CommissionMakeSettle(models.TransientModel):
         selection_add=[("sale_invoice", "Sales Invoices")],
         ondelete={"sale_invoice": "cascade"},
     )
+    date_payment_to = fields.Date(
+        "Payment date up to",
+        help="For payment-based commissions, settlements will be created for payments \
+            with date up to the one set in this field.",
+        default=fields.Date.today,
+    )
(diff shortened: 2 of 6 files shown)


account_commission/models/account_move.py
```diff
@@ -251,7 +251,30 @@ def _skip_settlement(self):
         :return: bool
         """
         self.ensure_one()
+        payment_based_commission = (
+            True if self.commission_id.invoice_state == "paid" else False
```
PicchiSeba: **nitpick**: this expression could be simplified

```suggestion
            self.commission_id.invoice_state == "paid"
```

PicchiSeba (changes requested):
Code review

odooNextev (approved):
LGTM

account_commission/wizards/commission_make_settle.py
```diff
@@ -46,3 +52,9 @@ def _prepare_settlement_line_vals(self, settlement, line):
                 }
             )
         return res
+
+    def action_settle(self):
+        context_date_payment = self.env.context.copy()
+        context_date_payment["date_payment_to"] = self.date_payment_to
+        self.env.context = context_date_payment
```
PicchiSeba: **question**: aren't these operations equivalent to

```py
self.env.context["date_payment_to"] = self.date_payment_to
```
?

Or am I missing something?

matteotognini: Hi @PicchiSeba, self.env.context is a frozendict, which is an immutable dictionary. Because of this, I couldn't add self.date_payment_to directly as you suggested.

PicchiSeba (approved):
Code review: LGTM

matteotognini:
@pedrobaeza What do you think? is it possible to merge?

account_commission/wizards/commission_make_settle.py
pedrobaeza: Please use `with_context` decorator instead of a hard replacement of this type, as it's not guaranteed that the internal implementation will work with this kind of operations.

pedrobaeza (changes requested):

PicchiSeba:
> Please use with_context decorator instead of a hard replacement of this type, as it's not guaranteed that the internal implementation will work with this kind of operations.

@pedrobaeza 

Shouldn't we prefer this approach instead of `with_context`? The old context would be discarded entirely

pedrobaeza:
No, `with_context` can have an entire dict as argument or just a context key. Depending on which you select, you will have a replacement or just a new/edited key.

account_commission/wizards/commission_make_settle.py
```diff
@@ -46,3 +52,7 @@ def _prepare_settlement_line_vals(self, settlement, line):
                 }
             )
         return res
+
+    def action_settle(self):
+        self = self.with_context(date_payment_to=self.date_payment_to)
+        return super().action_settle()
```
TheMule71, line 58: Is this covered by a test? 
My doubt is that `self` gets reassigned locally. I'm not sure that `super()` w/o arguments uses local `self`. So I'm not sure it works.

I'd suggest the explicit syntax:

`return super(CommissionMakeSettle, self).action_settle()`

to make sure that it's the modified `self` that gets used.

matteotognini: Hi @TheMule71, thanks for your review. I've just added a coverage test and it seems to work properly as it is now. Do you think it's better to make the syntax more explicit, as you mentioned? Thanks!

TheMule71: No, it's not necessary.

matteotognini:
@pedrobaeza merge?

pedrobaeza:
This is adding a bit of complication on the module for a specific case, but it's not too much overhead and it's tested, so let's continue.

/ocabot merge minor

Issue #153, opened 2018-04-21
commission should be deducted from margin


ferdiga:
![image](https://user-images.githubusercontent.com/6248560/39089350-67975354-45c5-11e8-913b-e3c0f1535ad2.png)

pedrobaeza:
For that, you will need to develop a bridge module `sale_commission_margin`, as it's not covered. Any way, some people will argue that the don't want to consider the commission as a margin reduction, but other kind of cost.

pedrobaeza:
Any way, this is not a bug itself, but a missing feature/module, so I close.

PR #591, opened 2025-01-27
[16.0] [ADD] sale_commission_margin: Deduct commission from margin

luisDIXMIT:
This addon allows commissions to be deducted from the margin.

Because precomputed fields must depend on other precomputed fields, some computed fields in the commission module have been established as precomputed.

Issue #153


commission/models/commission_mixin.py
```diff
@@ -19,13 +19,15 @@ class CommissionMixin(models.AbstractModel):
         readonly=False,
         store=True,
         copy=True,
+        precompute=True,
```
pedrobaeza: I don't think these changes are correct.

etobella: This changes are necessary in order to avoid warnings. The problem comes from the fact that margin fields are pre-computed, then it needs pre-computed fields.

However, after checking Odoo documentation, we can use pre-compute to fill this fields, as they are generated in batch (they are lines). This is why we thought it was ok.

https://www.odoo.com/documentation/16.0/developer/reference/backend/orm.html?highlight=precompute

If you have a reason that we cannot see (it might be possible), we would be happy if you can enlighten us.

pedrobaeza: Then change this attribute in the module `sale_commission_margin`, not here. I prefer to not change it due to things like `Precomputing a field can be counterproductive if the records of the given model are not created in batch`

etobella: I understand your concerns, but the following sentences states:

> Following the remark above, precomputed fields can be interesting on the lines of a one2many, which are usually created in batch by the ORM itself, provided that they are created by writing on the record that contains them

Usually, invoice_lines and sale order lines are generated with `[(0, 0, {}), ...]`. That means that they shouldn't be affected by the issue that you stated. 

I prefer to do it directly on the base model in order to avoid future errors or incompatibilities (if we add fields that are not pre-computed, warnings will be raised if this module is installed). However, if there are not other options, we can do it as you suggest.

Also, it is recommended to change the priority of the view on sale.order in order to show the commission and then the margin. Should we do that directly on this new module or can we do it on `sale_commission`?

pedrobaeza: Uhm, I'm still a bit reluctant. We don't know how can it impact. If precompute mode was so good, then it would be enabled by default. I still think that should be in the new module.

About the view priority switch, it can also be done on the new module. That way, you identify the whys of that changes.

etobella: It was made False because it only has sense in some models (mainly One2many fields) and it is problematic if we create items one by one.

Well, at this point I suggest to wait 24 hours if someone else makes its opinion, otherwise I will make the change directly in the module with the risks that it might imply.

etobella: @pedrobaeza We will pass the logic to the module, however, can we propose the change directly to 18 as it has not been migrated? I think it is much cleaner. I understand that it might be problematic on current installed versions as we don't know if it might affect.

pedrobaeza: But this is not only for current instances, but for any using that. The warning in the ORM definition is for every field using `precompute=True`. Why do you think it should be the default behavior?

etobella: Because of this:

https://github.com/odoo/odoo/blob/18.0/addons/sale/models/sale_order_line.py#L184-L203
https://github.com/odoo/odoo/blob/18.0/addons/sale_margin/models/sale_order_line.py#L13-L14

As the margin of the line is precomputed, we need to compute the amount of the commission on the line as precomputed in order to substract the amount.

Otherwise, it is not working properly, as the margin of the line is computed without the commission amount computed.

pedrobaeza: Yeah, I know that because you have already told me, but I mean for putting it on the main module. I think it should be still in your module in 18.0.

etobella: Why?

pedrobaeza: Because we don't know the impact and side effects, and that's not needed for the regular module to work. It's just to work (or at least, to not have the warning) with your extension module.

etobella: For that reason I am suggesting to use a new version. If we are going to block this kind of changes do we need to block all refactorings (even on migrations) too?

Also, by checking the Odoo documentation we showed that the solution makes sense. I don't understand why there are so many complains if we have provided answers with the issue and we have proposed a solution that minimizes effects on the current versions. Keeping it in the module might have other undesired effects as failing integrations with other commission modules and warnings. This is the reason why we are asking for the integration on the main module (we can accept to do it on a non released module).

Just to end, it is not "my" module. It is an OCA module that is fixing a behavior that is not working as expected when we install commission and a core community odoo module. It was documented 7 years ago and no one has addressed it since today

pedrobaeza: This sentence in the ORM documentation is the key: `Precomputing a field can be counterproductive if the records of the given model are not created in batch`. Activating `precompute` is not needed unless you install `sale_margin` and this glue module. That's why I say that flag should go into the glue module, not on `commission` or `sale_commission` one.

I don't see either why it's a problem to put it in the glue module and why you want it to put on the base module if it's not going to have any benefit, but on contrary, it can be counterproductive as stated in the documentation.

etobella: And the next sentence states that it makes sense in O2M fields like sale order lines, purchase order lines or account move lines:

`Following the remark above, precomputed fields can be interesting on the lines of a one2many, which are usually created in batch by the ORM itself, provided that they are created by writing on the record that contains them.`

It is not counterproductive IMO. Right now, commission and sale_margin do not work properly between them. We are just fixing this behavior.

The other option is to write down that we need to exclude sale_margin when using commission, as it makes no sense

pedrobaeza: But you are needing the same this glue module for deducting the amount, so even if you put `precompute=True` in the base module, you will still need the glue module. That's why I think it should go all together, the change to `precompute=True`, the view change, and the business logic.

etobella: And then we are leaving a big hole for possible future problems. 

I could understand your position when we are struggling with my own modules or OCA modules, but we are talking about an integration with a core community module of Odoo.

This are small changes that could be done on the commission module and we both know it.

etobella:
@OCA/crm-sales-marketing-maintainers

sale_commission_margin/models/commission_mixin.py
```diff
@@ -0,0 +1,50 @@
...
+        comodel_name="commission.line.mixin",
+        inverse_name="object_id",
+        string="Agents & commissions",
+        help="Agents/Commissions related to the invoice line.",
+        compute="_compute_agent_ids",
+        readonly=False,
+        store=True,
+        copy=True,
+        precompute=True,
+    )
```
pedrobaeza: You don't need to re-declare whole field:

```suggestion
    # convert existing fields to precompute
    agent_ids = fields.One2many(precompute=True)
```

luisDIXMIT: Hi @pedrobaeza, thanks for your review, but I have tried doing it this way and I get the following error. Is that okay anyway?

` warnings.warn(f"precompute attribute doesn't make any sense on non computed field {self}")`

pedrobaeza: OK, then try this, as the warning is not very fine-tuned:

```suggestion
    # convert existing fields to precompute
    agent_ids = fields.One2many(compute="_compute_agent_ids", precompute=True)
```

luisDIXMIT: Thanks, Pedro. I have added your suggestion, but I also had to add the store attribute to avoid this warning:
`warnings.warn(f"precompute attribute has no impact on non-stored field {self}").`

On the other hand, I have added a separate commit with a change to make the deduction information clearer by displaying the commission amount before the margin. This way, I prioritized the view, and now it looks as shown in the following image. Additionally, the commission information in the sale order lines is now displayed before the taxes and amount information. Previously, it was the other way around.

image

sale_commission/views/sale_order_view.xml
```diff
@@ -4,6 +4,8 @@
         <field name="name">sale.agent.order.inherit.form</field>
         <field name="model">sale.order</field>
         <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="priority" eval="15" />
+        <field name="mode">primary</field>
```
etobella: Why primary?

You should add the change directly in your module

luisDIXMIT: Because in that case, it has no effect. I looked into the Odoo documentation and supposed it is due to the following:

> View matching
> if a view is requested by (model, type), the view with the right model and type, mode=primary and the lowest priority is matched.
> 
> when a view is requested by id, if its mode is not primary its closest parent with mode primary is matched.

pedrobaeza: This is not correct. Please don't change it.

luisDIXMIT: Due to your suggestions and after several attempts with different approaches, I think the cleanest solution to show the commission total before the margin, without completely changing the view logic and while keeping it within my module, is to use the position attribute to move the field to a different node. WDY?

luisDIXMIT: I have also squashed all commits into one because it is all about my module.

sale_commission_margin/views/sale_order_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <record id="view_order_margin_form_inherit" model="ir.ui.view">
+        <field name="name">sale.order.margin.view.form</field>
+        <field name="model">sale.order</field>
+        <field name="inherit_id" ref="sale.view_order_form" />
+        <field name="arch" type="xml">
```
etobella, line 9: You should add a priority 99 in order to make the move after other changes.

luisDIXMIT: Comments attended!

etobella (approved):
LGTM

pedrobaeza (approved):
/ocabot merge nobump

PR #537, opened 2024-06-18
[14.0][ADD] sale_commission_delivery_partner

toita86:
This module solves inconsistency problems between SO and invoices
regarding the selected agent commissioners.

This module will make sure to use the Delivery Address' agents instead

PR #593, opened 2025-01-28
[14.0][ADD] sale_commission_delivery_partner: Use Delivery Address Agents

Chionne27:
Superseeds https://github.com/OCA/commission/pull/537


francesco-ooops (approved):
Functional ok!

@pedrobaeza what do you think?

sale_commission_delivery_partner/models/account_move_line.py
```diff
@@ -0,0 +1,14 @@
...
+    _inherit = "account.move.line"
+
+    @api.depends("move_id", "move_id.partner_shipping_id")
+    def _compute_agent_ids(self):
+        self.agent_ids = False
+        for record in self.filtered(lambda x: x.move_id.partner_shipping_id):
+            if not record.commission_free and record.product_id:
+                record.agent_ids = record._prepare_agents_vals_partner(
+                    record.move_id.partner_shipping_id
+                )
```
HekkiMelody: **issue**: overwriting functions without calling `super()` is an anti-pattern and should be avoided at all costs, especially in Odoo where several different configurations of modules could be installed. In this case, `_compute_agent_ids` is redefined by multiple modules, and I'm not at all sure about what the final result would be (even if the tests are currently green).

There are indeed some extreme cases where it's unavoidable, and those cases should at least be clearly signposted with comments explaining how and why this choice was made.

I think there's a fairly clean solution here that would allow us to avoid having to overwrite the function:

1. in the `sale_commission` module, in `sale.commission.mixin`, introduce a hook function: `_get_partner_for_commission` or something like that. That function would return the correct partner to use to calculate commissions. In `sale.commission.mixin` that function would just raise a `NotImplementedError()`.
2. In the `sale_commission` module, in the `account.move.line` and `sale.order.line`models, implement the "concrete" function `_get_partner_for_commission`. They would return `self.move_id.partner_id` and `self.order_id.partner_id` respectively.
3. In the `sale_commission` module, make a small change to `_compute_agent_ids` so that your new function `_get_partner_for_commission` is called instead of having `partner_id` hardcoded in the function. Do this also in other modules that override `_compute_agent_ids`.
4. In your new module, overwrite only `_get_partner_for_commission` to return `self.move_id.partner_shipping_id` and `self.order_id.partner_shipping_id`. You can still do a simple override of `_compute_agent_ids` to add the field you need to `api.depends` like this

```
    @api.depends("move_id.partner_shipping_id")
    def _compute_agent_ids(self):
        super()._compute_agent_ids()
```

What do you think?

sale_commission_delivery_partner/models/sale_order_line.py
```diff
@@ -0,0 +1,14 @@
...
+    _inherit = "sale.order.line"
+
+    @api.depends("order_id", "order_id.partner_shipping_id")
+    def _compute_agent_ids(self):
+        self.agent_ids = False
+        for record in self.filtered(lambda x: x.order_id.partner_shipping_id):
+            if not record.commission_free:
+                record.agent_ids = record._prepare_agents_vals_partner(
+                    record.order_id.partner_shipping_id
+                )
```
HekkiMelody: See other comment

HekkiMelody (changes requested):
Partial review. See comment

sale_commission_delivery_partner/models/account_move_line.py
Chionne27: Thank you for your detailed explanation. This approach actually enhances both maintainability and flexibility of code. Let me know if everything is right

Chionne27: Hi @aleuffre. Thank you fo the suggestion. This approach actually improves both maintainability and flexibility of code. Let me know if everything is right

HekkiMelody: I think the current implementation is missing point 4 and part of point 3.

PR #596, opened 2025-02-05
[16.0][IMP] sale_commission: Recompute agents amount button

christian-ramos-tecnativa:
The module commission_formula standard behavior allows for the assignment of commissions based on formulas, such as the number of agents involved in a transaction. However, an issue arises when agents commissions are added based on the total number of agents.
For example, consider the rule: 10% / n_agents. This means:
- If there is 1 agent, they receive 10%.
- If there are 2 agents, each receives 5%.
And so on.

The expected result:
- Agent1 = 2.5
- Agent2 = 2.5
- Agent3 = 2.5
- Agent4 = 2.5

The actual result:
- Agent1 = 10
- Agent2 = 5
- Agent3 = 3.33
- Agent4 = 2.5

This PR solve the issue recomputing the amount for every agent on any agent line change

### DIFF
--- a/sale_commission/views/sale_order_view.xml
+++ b/sale_commission/views/sale_order_view.xml
@@ -35,6 +35,12 @@
                         string="Regenerate agents"
                         states="draft,sent"
                     />
+                    <button
+                        name="recompute_lines_agents_amount"
+                        type="object"
+                        string="Recompute amounts"
+                        states="draft,sent"
+                    />
                 </group>
             </xpath>
         </field>
--- a/sale_commission/models/sale_order.py
+++ b/sale_commission/models/sale_order.py
@@ -42,6 +42,9 @@ class SaleOrder(models.Model):
     def recompute_lines_agents(self):
         self.mapped("order_line").recompute_agents()
 
+    def recompute_lines_agents_amount(self):
+        self.mapped("order_line").agent_ids._compute_amount()
+
 
 class SaleOrderLine(models.Model):
     _inherit = [
--- a/sale_commission/tests/test_sale_commission.py
+++ b/sale_commission/tests/test_sale_commission.py
@@ -136,6 +136,10 @@ class TestSaleCommission(TestAccountCommission):
         sale_order.recompute_lines_agents()
         agent = sale_order.order_line.agent_ids
         self._check_propagation(agent, self.commission_net_invoice, self.agent_monthly)
+        # Check recomputation of amount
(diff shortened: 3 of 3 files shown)


pedrobaeza (changes requested):
This is a very specific case that shouldn't be handled in the base module, because this provokes an unneeded overload for the usual cases, which involves just a computation not based on the total number of agents. You can do this in your custom end or in an extension module, but not here.

christian-ramos-tecnativa:
How about adding a new field at the commission type level, allowing users to specify whether this recomputation is needed? From my perspective, this serves more as a fix to ensure the module fully accommodates all possible scenarios and not sure if it make sense in a module separate to make this one work properly.
As an alternative we could create a new button at sale level called "Recompute agents amounts" next to the "Regenerate agents" button instead of changing the value in the line to trigger the recompute.

pedrobaeza:
Having the configuration option will provoke a similar overhead having to access it on each agent change. I think it's acceptable that the button "Regenerate agents" recomputes the amounts too as an extra (changing the help to indicate it).

christian-ramos-tecnativa:
I don’t think that’s a real solution since the "Regenerate Agents" button removes and recreates agents, also deleting any manually inserted ones. Instead, the recompute amount feature could be useful in other scenarios, such as updating commission type values and applying them to a pending quotation. Rather than triggering the computation by modifying values, users could leverage this new feature to achieve the same result more efficiently. Also could be added that the button is displayed just when there is at least one formula type agent.

pedrobaeza:
OK, let's go with the second button, but try to not overhead the UI.

christian-ramos-tecnativa:
Ready, tell me what you think about this approach please

sale_commission/views/sale_order_view.xml
```diff
@@ -35,6 +35,12 @@
                         string="Regenerate agents"
                         states="draft,sent"
                     />
+                    <button
+                        name="recompute_lines_agents_amount"
+                        type="object"
+                        string="Recompute Amounts"
```
pedrobaeza: Let's not reproduce orthographic errors (https://icaltefl.com/capital-letters-in-english):

```suggestion
                        string="Recompute amounts"
```

christian-ramos-tecnativa: Done

pedrobaeza (approved):

rvalyi (approved):
LGTM

PR #601, opened 2025-03-18
[16.0][IMP] account_commission: add contact_by_agent_ids method

antoniospneto:
This method returns a comma-separated list of agent IDs associated with the invoice, useful for placeholders in email templates.

### DIFF
--- a/account_commission/models/account_move.py
+++ b/account_commission/models/account_move.py
@@ -121,6 +121,13 @@ class AccountMove(models.Model):
         ).write({"state": "settled"})
         return super().unlink()
 
+    def contact_by_agent_ids(self):
+        """
+        Return a comma-separated string of agent IDs associated with the invoice.
+        Useful as a placeholder in the email template (e.g., for the partner_to field).
+        """
+        return ",".join(map(str, self.partner_agent_ids.ids))
+
 
 class AccountMoveLine(models.Model):
     _inherit = [
--- a/account_commission/tests/test_account_commission.py
+++ b/account_commission/tests/test_account_commission.py
@@ -59,31 +59,29 @@ class TestAccountCommission(TestCommissionBase):
             limit=1,
         )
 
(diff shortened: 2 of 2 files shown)


account_commission/models/account_move.py
```diff
@@ -121,6 +121,13 @@ def unlink(self):
         ).write({"state": "settled"})
         return super().unlink()
 
+    def contact_by_agent_ids(self):
+        """
+        Returns a comma-separated string of agent IDs associated with the invoice.
```
rvalyi: We use imperative style in the OCA (no s: return)

antoniospneto: fixed!

pedrobaeza:
AFAIK, you can do that expression inside the placeholder.

antoniospneto:
> AFAIK, you can do that expression inside the placeholder

So, until v14, I always built the expression directly in the field as a placeholder. However, now in v16, I couldn’t get it to work at all. It seems like Jinja has become more strict, haha. I couldn’t do it directly, at least not in a simple way. The Jinja expression was getting so complex that I gave up and decided to create a function to help with it.

pedrobaeza:
AFAIK, v16 is even more powerful for such expressions, and don't even need to do a list comprehension. Why not using directly `partner_agent_ids`?

antoniospneto:
At first, it didn't work, but maybe I made a mistake somewhere. I'll double-check and get back to you with my feedback soon. For now, I'll leave the PR as a draft. Thanks!

antoniospneto:
@pedrobaeza

I managed to solve it here, but I really had to use list comprehension.

Previously I used it like this (v14): `${object.partner_agent_ids.ids|join(', ') }` (Mako)
Now I had to do it like this (v16): `{{ ', '.join([str(agent) for agent in object.partner_agent_ids.ids]) }}` (QWeb)

I'll close the PR, thanks!

PR #583, opened 2024-12-06
[18.0] [MIG] sale_commission: Migration to 18.0

Borruso:
Migration from 17.0 to 18.0

Issue: https://github.com/OCA/commission/issues/564

PR #582, opened 2024-12-06
[18.0] [MIG] account_commission: Migration to 18.0

Borruso:
Migration from 17.0 to 18.0

Issue: https://github.com/OCA/commission/issues/564

PR #577, opened 2024-11-12
[18.0] [MIG] commission: Migration to 18.0

moaadbourhim:

PR #605, opened 2025-03-20
[18.0][MIG] Migration and rename of module sale_commission

alan196:
This is the same PR as https://github.com/OCA/commission/pull/583

I just renamed the module from sale_commission to sale_commission_oca

This is required because Odoo name their official enterprise module as sale_commission, if you add this repository to a database with Odoo Enterprise it won't work.

Depends on:
- #577
- #582


alan196:
@moaadbourhim Could you help me to review 🙏

matteotognini (approved):

sebalix (commented):
@alan196 you put the module renaming in the `DON'T MERGE` commit by mistake, better to have it in its own commit.
I would do it this order:

1. rename the module before the migration:
`[MOV] Rename sale_commission to sale_commission_oca`
2. run pre-commit so README etc are updated according to the new module name:
`[IMP] sale_commission_oca: pre-commit execution`
3. migrate the module:
`[MIG] sale_commission_oca: Migration to 18.0`

JordiMForgeFlow:
@sebalix @alan196 should we maybe change also commission module to commission_oca, so it is aligned?

PR #602, opened 2025-03-20
[18.0] [MIG] commission: Migration to 18.0

alan196:
This PR supersedes https://github.com/OCA/commission/pull/577

It's the same PR applying the commit order in https://github.com/OCA/commission/pull/577#issuecomment-2472611849

PR #610, opened 2025-04-22
[18.0][MIG] commission + rename to commission_oca

JordiMForgeFlow:
Supersedes https://github.com/OCA/commission/pull/602 due to inactivity

sale_commission will need to be renamed to sale_commission_oca (https://github.com/OCA/commission/pull/605), so we could rename the base module to commission_oca for alignment.

CC @alan196


JordiMForgeFlow:
@pedrobaeza could we get some reviews here as authors?

JasminSForgeFlow (approved):
LGTM!

JordiMForgeFlow:
@sebalix could we get your input in this one?

sebalix (approved):
Don't know well this module, but from a technical POV it loogs good. I agree on the renaming too.

For info Odoo EE is proposing the following module technical names:
- `partner_commission`
- `sale_commission`
- `sale_commission_margin`
- `sale_commission_subscription`

pedrobaeza:
/ocabot migration commission

Thanks for the migration and the renaming. I'm afraid the renaming you have done doesn't include the whole commit history, so on the next version, only the last commit will be rescued with the procedure.

For avoiding this, you have to rename the whole history, please execute the following commands in your branch:

```
git filter-branch --tree-filter 'if [ -d commission ]; then mv commission commission_oca; fi' HEAD
git rebase origin/18.0
git push -f
```

and it will be done for the whole commit history.

Another thing: for the PO/POT, you have to also rename the module on each translation string comment.

commission_oca/i18n/it.po
```diff
@@ -18,7 +18,7 @@ msgstr ""
 
 #. module: commission
```
pedrobaeza: Here (and all the occurrences) you have to rename as well.

JordiMForgeFlow:
@pedrobaeza it should now be ready :) thank you!

pedrobaeza (approved):
Thanks!

/ocabot merge nobump

MiquelRForgeFlow:
Shouldn't be done anything in OpenUpgrade?

pedrobaeza:
Yes, the renaming should happen also in OpenUpgrade, so you can proceed with the PRs. Thanks for being aware.

PR #613, opened 2025-04-29
[FIX] account_commission: Use partner on the lang in order to avoid errors later with delegated partners

etobella:
If we try to invoice two agents with the same delegated partner, it fails


pedrobaeza (approved):

kluna1998 (approved):
👍🏽

PR #619, opened 2025-05-20
[16.0] [FIX] sale_commission_product_criteria_domain: call super on _get_commission_items

Tisho99:
The module sale_commission_product_criteria_domain is overriding _get_commission_items without calling super(). 

This causes trouble creating modules that depend on sale_commission_product_criteria and having sale_commission_product_criteria_domain installed.

In my opinion, the _get_commission_items function does not need to be overridden, because the functionality added in this module is bounded to the new commission type of this module, we can let the function work as before if the commission type is not equal to the one this module adds

T-8297

### DIFF
--- a/sale_commission_product_criteria_domain/models/sale_commission_line_mixin.py
+++ b/sale_commission_product_criteria_domain/models/sale_commission_line_mixin.py
@@ -8,59 +8,62 @@ class SaleCommissionLineMixin(models.AbstractModel):
     _inherit = "commission.line.mixin"
 
     def _get_commission_items(self, commission, product):
-        # Method replaced
-        categ_ids = {}
-        categ = product.categ_id
-        while categ:
-            categ_ids[categ.id] = True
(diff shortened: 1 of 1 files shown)


ValentinVinagre (changes requested):
One return please.

Tisho99:
> One return please.

Changed

ValentinVinagre:
> > One return please.
> 
> Changed

'res' is not defined in the function

Tisho99:
> > > One return please.
> > 
> > 
> > Changed
> 
> 'res' is not defined in the function

Changed

manuelregidor (approved):
LGTM

ValentinVinagre (approved):
LGTM 👍🏻

ValentinVinagre:
@pedrobaeza could u review ?

pedrobaeza:
Merging following current reviews: 

/ocabot merge patch

PR #626, opened 2025-07-16
[ADD] commission_oca: adding migrations

Borruso:


pedrobaeza (changes requested):
This seems incorrect and something about a jump from one old version to this one.

PR #628, opened 2025-07-22
[15.0][MIG] sale_commission_settlement_report: update inherit_id and manifest for v15

PolComas:
This PR makes minimal changes to bring sale_commission_settlement_report in line with Odoo 15:

- Correct the `inherit_id` in report_settlement_templates.xml so that it extends the proper QWeb template
- Update __manifest__.py:
  - Set version to 15.0.1.0.0
  - Add 'commission' to dependencies


PolComas:
The drop is only 0.03% due to an unrelated indirect file. Patch is 100%.
@OCA-git-bot overridecodecov

etobella (changes requested):
Follow migration guidelines:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

You have lost all history of commits. Also. The migration should have just one commit and you can use pre-commit to generate the setup automatically

PR #630, opened 2025-07-31
[15.0][IMP] commission: Add the ability to set commissions by product or category

CarlosRoca13:
cc @Tecnativa TT57153

ping @carlosdauden @sergio-teruel

PR #631, opened 2025-08-01
[15.0][MIG] sale_commission_product_criteria: migration to v15

PolComas:
Migration to 15.0


PolComas:
Hi @CarlosRoca13 could you please take another look at this PR now that I’ve restored history, squashed into one commit, and reverted the problematic onchange changes? I’d appreciate a final review & merge if everything looks good. Thanks!

sale_commission_product_criteria/views/views.xml
```diff
@@ -0,0 +1,229 @@
...
+                    >
+                        <tree>
+                            <field name="sequence" invisible="1" />
+                            <field name="name" string="Applied On" />
+                            <field name="commission_value" />
+                            <field name="based_on" invisible="1" />
+                        </tree>
+                    </field>
+                </group>
+            </form>
```
CarlosRoca13:
```suggestion
            <xpath expr="//field[@name='section_ids']/.." position="after">
                <group
                    string="Rules"
                    name="rules_group"
                    colspan="4"
                    attrs="{'invisible': [('commission_type', '!=', 'product')]}"
                >
                    <field
                        name="item_ids"
                        nolabel="1"
                        context="{'hide_commission_id':1}"
                        colspan="4"
                    >
                        <tree>
                            <field name="sequence" invisible="1" />
                            <field name="name" string="Applied On" />
                            <field name="commission_value" />
                            <field name="based_on" invisible="1" />
                        </tree>
                    </field>
                </group>
            </xpath>
```

CarlosRoca13 (commented):
Functional review :smile: 

Check the commision view :sweat_smile: 
image

CarlosRoca13:
ping @carlosdauden

sale_commission_product_criteria/views/views.xml
PolComas: Thanks for the feedback!

I ended up using an:`<xpath expr="//sheet" position="inside">…</xpath>"` instead of replacing the `<form>`, and now the Rules block sits correctly inside the frame.

If you’re happy with this, we can consider the point closed. Otherwise, let me know and I’ll switch it to target the section_ids node exactly.
Captura de 2025-08-01 11-43-17

CarlosRoca13 (approved):
Thanks :smile:

PolComas:
Hi @ilyasProgrammer @aleuffre  could you please take a look at this migration to 15.0 and merge if it looks good? Thanks!

sale_commission_product_criteria/models/sale.py
```diff
@@ -30,7 +30,7 @@ def _compute_amount(self):
                     order_line.product_uom_qty,
                 )
             else:
-                super(SaleOrderLineAgent, line)._compute_amount()
+                return super(SaleOrderLineAgent, line)._compute_amount()
```
pedrobaeza: You can't return in a loop.

sale_commission_product_criteria/__manifest__.py
```diff
@@ -3,13 +3,13 @@
...
     "summary": "Advanced commissions rules",
-    "version": "14.0.1.1.6",
+    "version": "15.0.1.0.0",
     "author": "Ilyas, Ooops404, Odoo Community Association (OCA)",
     "maintainers": ["ilyasProgrammer", "aleuffre", "renda-dev", "PicchiSeba"],
     "website": "https://github.com/OCA/commission",
     "category": "Sales Management",
     "license": "AGPL-3",
-    "depends": ["sale_commission", "web_domain_field"],
+    "depends": ["commission", "sale_commission", "web_domain_field"],
```
pedrobaeza: No need of changing this, as `sale_commission` already depends on `commission`.

sale_commission_product_criteria/models/account_move.py
```diff
@@ -14,18 +14,14 @@ class AccountInvoiceLineAgent(models.Model):
         "commission_id",
     )
     def _compute_amount(self):
+        res = super(AccountInvoiceLineAgent, self)._compute_amount()
```
pedrobaeza:
```suggestion
        res = super()._compute_amount()
```

pedrobaeza (changes requested):

sale_commission_product_criteria/models/account_move.py
PolComas: I have applied the changes you suggested, thanks for the review!

sale_commission_product_criteria/models/sale.py
pedrobaeza: This is still not changed.

sale_commission_product_criteria/models/account_move.py
```diff
@@ -14,18 +14,14 @@ class AccountInvoiceLineAgent(models.Model):
...
             if line.commission_id and line.commission_id.commission_type == "product":
                 inv_line = line.object_id
-                amount = line._get_single_commission_amount(
+                line.amount = line._get_single_commission_amount(
                     line.commission_id,
                     inv_line.price_subtotal,
                     inv_line.product_id,
                     inv_line.quantity,
                 )
-                if line.invoice_id.move_type == "out_refund":
```
pedrobaeza: Is the sign now handled in the method?

sale_commission_product_criteria/__manifest__.py
pedrobaeza: This is still not changed.

sale_commission_product_criteria/models/sale.py
```diff
@@ -0,0 +1,54 @@
...
+                order_line = line.object_id
+                line.amount = line._get_single_commission_amount(
+                    line.commission_id,
+                    order_line.price_subtotal,
+                    order_line.product_id,
+                    order_line.product_uom_qty,
+                )
+            else:
+                res = super(SaleOrderLineAgent, line)._compute_amount()
+        return res
```
CarlosRoca13, line 35:
```suggestion
    def _compute_amount(self):
        lines = self.env["sale.order.line.agent"]
        for line in self:
            if line.commission_id and line.commission_id.commission_type == "product":
                order_line = line.object_id
                line.amount = line._get_single_commission_amount(
                    line.commission_id,
                    order_line.price_subtotal,
                    order_line.product_id,
                    order_line.product_uom_qty,
                )
            else:
                lines += line 
        return super(SaleOrderLineAgent, lines)._compute_amount()
```
I believe the most appropriate way to achieve the goal is the following

sale_commission_product_criteria/models/account_move.py
CarlosRoca13: Please take this comment into account as well. The method _get_single_commission_amount is not handling the sign

sale_commission_product_criteria/models/sale_commission_line_mixin.py
```diff
@@ -0,0 +1,77 @@
...
+            """,
+            (
+                product.product_tmpl_id.ids,
+                product.ids,
+                categ_ids,
+                commission._origin.id,  # Added this
+            ),
+        )
+        item_ids = [x[0] for x in self.env.cr.fetchall()]
+        return item_ids
```
CarlosRoca13:
```suggestion
    def _get_commission_items(self, commission, product, _select="", _from="", _where="", _order_by=""):
        # Method replaced
        categ_ids = {}
        categ = product.categ_id
        while categ:
            categ_ids[categ.id] = True
            categ = categ.parent_id
        categ_ids = list(categ_ids)
        # Select all suitable items. Order by best match
        # (priority is: all/cat/subcat/product/variant).
        self.env.cr.execute(
            """
            SELECT
                item.id
                %s
            FROM
                commission_item AS item
            LEFT JOIN product_category AS categ ON item.categ_id = categ.id
                %s
            WHERE
                (item.product_tmpl_id IS NULL OR item.product_tmpl_id = any(%s))
                AND (item.product_id IS NULL OR item.product_id = any(%s))
                AND (item.categ_id IS NULL OR item.categ_id = any(%s))
                AND (item.commission_id = %s)
                AND (item.active = TRUE)
                %s
            ORDER BY
                item.applied_on, item.based_on, categ.complete_name desc
                %s
            """,
            (
                AsIs(_select),
                AsIs(_from),
                product.product_tmpl_id.ids,
                product.ids,
                categ_ids,
                commission._origin.id,  # Added this
                AsIs(_where),
                AsIs(_order_by),
            ),
        )
        item_ids = [x[0] for x in self.env.cr.fetchall()]
        return item_ids
```

Let's make the method more extensible :smile:

CarlosRoca13 (changes requested):

sale_commission_product_criteria/models/sale_commission_line_mixin.py
pedrobaeza: Can't this be replaced by a ORM search?

carlosdauden: For this purpose a related field must be added to categ_id.complete_name

CarlosRoca13: And add this new field to the `_order` of `commission.item` model

carlosdauden: It was the option used here:
https://github.com/OCA/commission/pull/630/files#diff-bf37eeffd58715fd99bfbd5f6c2149166deb2a13a0f73cd2aea98f72a0143f6dR97****

PolComas: For now, I would keep the current implementation, as it is clear, self-contained, and working as expected. If there is a strong preference in OCA for the ORM/related field approach, I am open to reconsidering, but I think this solution is also valid.

sale_commission_product_criteria/models/account_move.py
PolComas: Yes, the sign is now handled inside _compute_amount.
In case of a refund (out_refund), the amount is explicitly negated.
I also added a unit test (test_commission_sign_on_customer_refund) to ensure that the commission on the refund is always the negative of the original.

sale_commission_product_criteria/__manifest__.py
PolComas: Changes applied👍

sale_commission_product_criteria/models/sale.py
PolComas: Changes applied👍

sale_commission_product_criteria/models/sale.py
CarlosRoca13: Why this change is not set?

sale_commission_product_criteria/models/sale_commission_line_mixin.py
CarlosRoca13: Alright, but could you at least add the change I’m suggesting so the method is more extensible? We need it for this PR: https://github.com/OCA/commission/pull/634

PolComas: Changes applied 🥳

pedrobaeza: > For now, I would keep the current implementation, as it is clear, self-contained, and working as expected. If there is a strong preference in OCA for the ORM/related field approach, I am open to reconsidering, but I think this solution is also valid.

@PolComas SQL is ignoring the record rules you have applied over the system, so it's far from ideal. For example, the SQL doesn't have any company consideration right now, mixing entries from several companies in the result.

CarlosRoca13 (approved):
Thanks!

pedrobaeza (changes requested):
For me, the SQL is still blocking.

ivilata:
This PR has been superseded by #636 by @CarlosRoca13. @PolComas, would you mind closing this one to reduce clutter? Thanks! :slightly_smiling_face:

PR #636, opened 2025-08-26
[15.0][MIG] sale_commission_product_criteria: Migration to 15.0

CarlosRoca13:
This PR supersedes https://github.com/OCA/commission/pull/631 to replace the use of `SQL` with a `filtered_domain` on the commission items. The authorship of the migration commit has been preserved.

cc @Tecnativa TT57153

ping @carlosdauden @pedrobaeza @PolComas


sale_commission_product_criteria/models/commission.py
```diff
@@ -65,7 +65,7 @@ def check_type_change_allowed_moves(self):
 class CommissionItem(models.Model):
     _name = "commission.item"
     _description = "Commission Item"
-    _order = "applied_on, based_on, categ_id desc, id desc"
+    _order = "applied_on, based_on, category_complete_name desc, id desc"
```
pedrobaeza: I don't think you need this change, but change `_order` of the category itself.

sale_commission_product_criteria/models/sale_commission_line_mixin.py
```diff
@@ -19,58 +18,43 @@ class SaleCommissionLineMixin(models.AbstractModel):
...
 
     def _get_single_commission_amount(self, commission, subtotal, product, quantity):
         self.ensure_one()
-        item_ids = self._get_commission_items(commission, product)
-        if not item_ids:
+        items = self._get_commission_items(commission, product)
+        if not items:
             return 0.0
-        commission_item = self.env["commission.item"].browse(item_ids[0])
+        commission_item = items[0]
```
pedrobaeza: Why not limiting search to 1?

sale_commission_product_criteria/models/commission.py
CarlosRoca13: Do you think it's more convenient to change the order of the category than to add this order at the `commission.item` level?

pedrobaeza: But why the default category order is not valid? It should be...

CarlosRoca13: As I understand it, adding a many2one field in the order evaluates the ID of the related record, but what we need is for it to be ordered by grandchild > child > parent. If we create the child record first, then the parent, and then the grandchild, the order would be incorrect (grandchild > parent > child). By ordering by complete_name (as categories do), we solve the problem. Am I wrong?

sale_commission_product_criteria/models/sale_commission_line_mixin.py
CarlosRoca13: Done

sale_commission_product_criteria/models/commission.py
pedrobaeza: When adding a many2one, first you get the IDs ordered by their native model order, and then it's applied here, so you don't need to reorder them (this also means 2 queries, but Odoo does it this way for UX convenience).

CarlosRoca13: Thank you for the clarification

carlosdauden: This change leads to worse performance at a rather delicate moment, but if you want we can leave it for the time being until we can measure it in an instance with a lot of records.

sale_commission_product_criteria/models/commission.py
```diff
@@ -0,0 +1,265 @@
+# © 2023 ooops404
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.tools import float_repr
+
+
+class SaleCommission(models.Model):
```
pedrobaeza:
```suggestion
class Commission(models.Model):
```

sale_commission_product_criteria/models/commission.py
pedrobaeza: Is it the number of records to be very high for the commission definition? Do you think it has a big impact? Other thing it can be done is to put `auto_joint=True` on `categ_id` for resolving everything in one query.

pedrobaeza (approved):

CarlosRoca13:
ping @carlosdauden

carlosdauden (approved):

PR #651, opened 2025-11-17
[16.0][IMP] account_commission: set negative amount when the invoice is a Customer Credit Note

CarlosRoca13:
cc @Tecnativa TT59290

ping @pedrobaeza @carlosdauden


pedrobaeza (approved):
Let's fast-track it as already merged in previous and being pretty obvious.

/ocabot merge ptch

PR #653, opened 2025-12-11
[18.0][MIG] sale_commission_product_criteria_oca: Migration to 18.0

FranzPoize:
I know this is a bit rude to create another PR for this migration but I kinda need this to be merged quickly and the PR seems dead for a month.

Module is migrated, tests are running and functionally I tested on a small case and it seems to work.

The module name is changed to reflect the name change of other modules.


FranzPoize:
Can someone review this ? @Tisho99 @ilyasProgrammer 
Thanks in advance

FranzPoize:
It would be great if someone could review this please

PR #655, opened 2026-01-22
[16.0][FIX] account_commission: Doesn't compute data upon installation

Shide:
Skip initial computation of account_move and account_move_line problematic fields.

MT-12373 @moduon @rafaelbn @pedrobaeza @yajo please review if you want 😄


pedrobaeza (commented):
I think it's OK to add this improvement for speeding up installation (although it shouldn't be very heavy). Please initialize the values to the proper ones (0 an False instead of `NULL`).

Shide:
> I think it's OK to add this improvement for speeding up installation (although it shouldn't be very heavy). Please initialize the values to the proper ones (0 an False instead of `NULL`).

done :)

pedrobaeza (approved):

carlosdauden (approved):

PR #658, opened 2026-01-23
[FIX][16.0] sale_commission: UX improvements

Shide:
UX improvements:
- Commission actions are now in one line with and visible to Commission users: Recomputation of prices are triggered automatically, so, no need to allow anyone to restore agents if a Commission Manager added a new one. Regenerate agents button are only allowed to Commission Managers (the one who could add more agents)

MT-12373 @moduon @rafaelbn @pedrobaeza @yajo @Gelojr @fcvalgar please review if you want 😄

### DIFF
--- a/sale_commission/views/sale_order_view.xml
+++ b/sale_commission/views/sale_order_view.xml
@@ -28,19 +28,28 @@
                 />
             </field>
             <xpath expr="//field[@name='note']" position="after">
-                <group name="group_recompute_commission">
-                    <button
-                        name="recompute_lines_agents"
-                        type="object"
-                        string="Regenerate agents"
-                        states="draft,sent"
-                    />
-                    <button
-                        name="recompute_lines_agents_amount"
-                        type="object"
-                        string="Recompute amounts"
-                        states="draft,sent"
-                    />
+                <group
+                    name="group_recompute_commission"
+                    attrs="{'invisible': [('state', 'not in', ['draft', 'sent'])]}"
+                    colspan="2"
+                    groups="commission.group_commission_user"
+                    string="Commission actions"
+                >
+                    <div class="o_row" colspan="2">
+                        <button
+                            name="recompute_lines_agents"
+                            type="object"
+                            string="Regenerate agents"
+                            class="btn-secondary me-3"
+                            groups="commission.group_commission_manager"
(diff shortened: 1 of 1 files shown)


sale_commission/views/sale_order_view.xml
```diff
@@ -11,14 +11,16 @@
...
+                <div groups="commission.group_commission_manager">
+                    <button
+                        name="button_edit_agents"
+                        title="Edit agents"
+                        icon="fa-users"
+                        attrs="{'invisible': ['|', ('commission_free', '=', True), ('invoice_lines', '!=', [])]}"
+                        type="object"
+                    />
+                </div>
+                <field name="commission_status" optional="hide" />
```
pedrobaeza: Please keep it "show", as this change of behavior is a bit opinionated, and current users expect to see it, and those not wanting it, can still hide it.

sale_commission/views/sale_order_view.xml
```diff
@@ -11,14 +11,16 @@
...
                 <field name="invoice_lines" invisible="1" />
-                <button
-                    name="button_edit_agents"
-                    title="Edit agents"
-                    icon="fa-users"
-                    attrs="{'invisible': ['|', ('commission_free', '=', True), ('invoice_lines', '!=', [])]}"
-                    type="object"
-                />
-                <field name="commission_status" optional="show" />
+                <div groups="commission.group_commission_manager">
```
yajo: question: would it make more sense to add `groups` to the button directly?

pedrobaeza (commented):
About the security groups, I need to check also the implications, but these days I'm in a travel.

yajo (approved):
A couple of screenshots of before and after would help 😊 

Code review only.

sale_commission/views/sale_order_view.xml
Shide: Groups on button doesn't work. This is why I've done this, but it's breaking some things.
I need to work more on this to fix integrations with other modules.

Shide:
I've left only the change in recompute commission actions. Other changes will go in a separate PR because implies several modules.

Shide:
Closed in favor of https://github.com/OCA/commission/pull/659

PR #663, opened 2026-03-03
[19.0][IMP] commission_oca: add agents to partner list view

JordiMForgeFlow:


pedrobaeza:
Any explanation about this?

JordiMForgeFlow:
@pedrobaeza just adding the field in the list view for visibility and usability for bulk update in the list view

pedrobaeza:
Uhm, this means to reduce the performance, as the relation table should be linked and displayed. Have you measured that? I like the mass-edition thing, but we shouldn't affect too much the performance of this critical view.

JordiMForgeFlow:
@pedrobaeza I am hiding the field by default, so it should not have that much impact right?

pedrobaeza:
Well, no, although you put `optional="hide"`, the field is populated.

JordiMForgeFlow:
@pedrobaeza do you have any recommendations on how to measure the impact of the change to evaluate if we can add it in the base module or not?

pedrobaeza:
On a DB with populate tool to have a lot of partners, you should populate also agents (it can be a good moment to add the sql to populate them).

After that, open a profiling without that field, and showing a generous amount of partners in one shot, and do the same with the field. Then, comparing the profiling, you can measure the impact in percentage of difference.

JordiMForgeFlow:
@pedrobaeza I tested with a DB of 50000 partners and opening in one shot 10000 partners:

image

the resulting experience was very similar. Both scenarios were actually a bit slow when dealing with that amount of records. On another note, what we are adding is pretty much the same as it is happening by default with the partner tags.

pedrobaeza:
Well, all is additive, so 3 seconds more adding that field is a 20% more of time, so it's not negligible. Let's do this: create a new menu inside "Commission" section with a dedicate tree view with that multi_edit="1" attribute, and the fields you consider (name and agents, but maybe any other...), and this way, we have the best of both worlds: no penalty for usual contact view, and a way to change this massively when needed.

JordiMForgeFlow:
@pedrobaeza I followed your suggested approach, it is ready for review :)

commission_oca/views/res_partner_views.xml
```diff
@@ -70,6 +70,73 @@
...
+                    groups="base.group_multi_company"
+                    readonly="1"
+                    optional="hide"
+                />
+                <field name="properties" optional="hide" />
+            </list>
+        </field>
+    </record>
+    <record id="action_partner_commission" model="ir.actions.act_window">
+        <field name="name">Customers</field>
```
pedrobaeza:
```suggestion
        <field name="name">Customers with agents</field>
```

commission_oca/views/res_partner_views.xml
```diff
@@ -84,4 +151,11 @@
         action="action_agent_form"
         sequence="10"
     />
+    <menuitem
+        id="menu_partner_commission"
+        name="Customers"
```
pedrobaeza:
```suggestion
        name="Customers with agents"
```

commission_oca/views/res_partner_views.xml
```diff
@@ -70,6 +70,73 @@
...
+        <field name="name">Customers</field>
+        <field name="type">ir.actions.act_window</field>
+        <field name="res_model">res.partner</field>
+        <field name="view_mode">list,form</field>
+        <field name="domain">[('parent_id', '=', False), ('agent', '=', False)]</field>
+        <field
+            name="view_ids"
+            eval="[(5, 0, 0), (0, 0, {'view_mode': 'list', 'view_id': ref('view_partner_list_commission')})]"
+        />
+        <field name="search_view_id" ref="res_partner_view_search" />
```
pedrobaeza, line 145: Put a default context for filtering only partners with agents.

JordiMForgeFlow:
@pedrobaeza done! I add a default filter to allow the user to end up seeing all customers in case they want to use the view to mass assign agents to customers that don't currently have one.

pedrobaeza (approved):

PR #665, opened 2026-03-05
[19.0][MIG] hr_commission_oca

omar7r:
Standard migration


hr_commission_oca/views/res_partner_view.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record model="ir.ui.view" id="view_partner_form_agent">
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="commission_oca.view_partner_form_agent" />
+        <field name="arch" type="xml">
+            <field name="commission_id" position="after">
+                <field name="employee" invisible="agent_type != 'salesman'" />
```
pedrobaeza, line 8: This field has been removed, isn't it?

carlos-lopez-tecnativa: As I understand it, in V18 the field was a regular field in the base module, which is why this module converted it into a computed field.

Now, in V19, Odoo added this field as a computed field in this PR: https://github.com/odoo/odoo/pull/181782. For this reason, I understand why @omar7r removed the field from the model but kept it in the view.

However, I noticed a few differences that probably mean we need to update the code.

We rely on [user_ids](https://github.com/OCA/commission/blob/288b2a8a62920657b40b845983fcc31860463af6/hr_commission_oca/models/res_partner.py#L25) having employees; in other words, we expect employees to have a user, whereas Odoo now relates employees to partners without necessarily having a user. See https://github.com/odoo/odoo/blob/615e54ecd2722433953931e1d51be15b069288c3/addons/hr/models/res_partner.py#L11.

I noticed that this module does not use `employee` or `employee_id` in any of its logic. These fields could be useful if we add an extension to integrate commissions with payroll, but until then, are these fields really necessary, @pedrobaeza?

hr_commission_oca/i18n/ca_ES.po
pedrobaeza, line 1: This file should be integrated into `ca.po` (if any missing term) and removed.

hr_commission_oca/i18n/hr_commission.pot
pedrobaeza, line 1: This file should be removed.

hr_commission_oca/static/description/icon.svg
pedrobaeza, line 1: Wasn't the main icon changed?

hr_commission_oca/views/res_partner_view.xml
pedrobaeza: Yes, I think we can remove `employee_id` and just use boolean `employee` for now.

pedrobaeza (changes requested):

carlos-lopez-tecnativa:
Superseded by https://github.com/OCA/commission/pull/685

pedrobaeza:
As the other PR is preserving attribution, let's switch for finishing this.

PR #669, opened 2026-03-31
[14.0][FIX] sale_commission_product_criteria: address same label warning

HekkiMelody:
Fixes the following warning

```
2026-03-31 12:14:35,919 1 WARNING devel odoo.addons.base.models.ir_model: Two fields (commission_type, commission_id) of commission.item() have the same label: Commission Type. 
```

### DIFF
--- a/sale_commission_product_criteria/models/commission.py
+++ b/sale_commission_product_criteria/models/commission.py
@@ -71,7 +71,6 @@ class CommissionItem(models.Model):
     active = fields.Boolean(default=True)
     commission_id = fields.Many2one(
         "sale.commission",
-        string="Commission Type",
         domain=[("commission_type", "=", "product")],
         required=True,
     )


HekkiMelody:
Pre-commit failure addressed in https://github.com/OCA/commission/pull/670

sale_commission_product_criteria/models/commission.py
```diff
@@ -119,6 +119,7 @@ class CommissionItem(models.Model):
     )
     commission_type = fields.Selection(
         [("fixed", "Fixed"), ("percentage", "Percentage")],
+        string="Type",
```
SirPyTech: Before this change both `commission_type` and `commission` are labeled `Commission Type`.
After this change `commission_type` is labeled `Type` and `commission` is labeled `Commission Type`.
I think the new behavior is just as confusing as the old one.

Maybe the real problem is a few lines above:
```py
    commission_id = fields.Many2one(
        "sale.commission",
        string="Commission Type",
```
why don't we:
- label `commission_id` as `Commission`
- label `commission_type` as `Commission Type`

?

This would also be aligned with `18.0`, see https://github.com/OCA/commission/blob/2fefe980e5d75be4621792ab9168cf07c8a760b7/sale_commission_oca_product_criteria/models/commission.py#L73-L78 and https://github.com/OCA/commission/blob/2fefe980e5d75be4621792ab9168cf07c8a760b7/sale_commission_oca_product_criteria/models/commission.py#L120-L125.

In order to achieve this result, we can just remove `string` from both fields, or change `Commission Type` to `Commission` as it is in `18.0`.
What do you think?

SirPyTech (changes requested):

sale_commission_product_criteria/models/commission.py
HekkiMelody: Removed `string` attribute from both fields (fewer lines of code is always good!) to adhere to the same nomenclature as v18.

Thank you for the suggestion.

SirPyTech: :+1:

SirPyTech (approved):

SirPyTech:
Rebased to include https://github.com/OCA/commission/pull/670.

francesco-ooops (approved):
LGTM

ivs-cetmix:
Hi @HekkiMelody , thank you for your contribution!
/ocabot merge patch

PR #676, opened 2026-05-13
[16.0] [FIX] sale_commission_agent_restrict: remove the contact pages for agents

rrebollo:
## What changed

A contact has two form pages that agent users must not see: `Sales & Purchases` and
`Internal Notes`. They were hidden with a `groups` attribute on the pages themselves. This
PR keeps the same behaviour — the pages are removed for agent users — but removes them while
the form is being built, instead of relying on that attribute.

## Why

In v16, `groups` on a view node does not merely hide it: the node **and every field inside
it** are removed from the architecture that is stored and validated. That breaks any other
module that knows nothing about this one. Such a module can add a field inside one of those
pages and then use it in the `attrs` or `domain` of a node that lives outside the page. The
field is then declared under a group restriction but used as mandatory for everyone, and
Odoo refuses to load the view:

```
odoo.tools.convert.ParseError: while parsing .../views/....xml:...
Error while validating view near:

<form string="Partners" __validate__="1"> ... </form>

Field '<field>' used in attrs ({'invisible': [('<field>', '=', False)]})
is restricted to the group(s) !sale_commission_agent_restrict.group_agent_own_commissions.
```

Updating the module aborted on exactly that error. The order matters: inherited views are
applied **before** the group restriction, so the third-party node is built while the page
still carries `groups`, and the conflict only surfaces when the view is validated. Putting
`groups` on a page is not a workable way to hide it from a group.

## How it works

The `groups` attributes are gone from the stored view, so the stored architecture is never
restricted and no third-party view can be invalidated by this module.

The pages are removed while the form architecture is built, only for agent users. The result
of that build is what gets cached, so the membership is part of what identifies the cached
version: one version for agents and one for everybody else, and nothing extra to pay each
time a contact form is opened. Two consequences are worth noting:

- The pages are gone for agents and present for everyone else, and the fields other modules
  need from those pages (`user_id`, `team_id`) are still available to agents through the
  hidden duplicates this module already shipped.
- Because the removal now happens earlier than it used to, the fields of those pages are not
  part of the form at all for an agent, so they are no longer sent to the browser — which is
  the point of removing the pages in the first place.

## Evidence

- Before the change, updating the module aborted with the `ValidationError` shown above.
  After the change the same update completes.
- Against 16.0, an agent gets the same pages removed and the same remaining fields. The
  differences are the intended ones: the fields of those pages are no longer part of the form
  (and, as a consequence, a client-side recomputation trigger that referred to them is no
  longer emitted).
- Three regression tests were added to `tests/test_sale_commission_agent_restrict.py`, reusing
  the fixtures that module already had:
  - one creates the offending view at runtime — a field inside the page, used in an `attrs`
    outside it — and asserts it can be created. It fails on 16.0 with the error above.
  - one asserts the pages are present for a non-agent and absent for an agent, with a cold
    cache and the non-agent loading first. It fails if the per-user decision is ever dropped
    from what identifies the cached version.
  - one asserts that agent users can still open the contact forms that never had those pages
    (simple, address and private forms, plus the `force_email` context path).

## Alternatives considered

- Keep `groups` and duplicate every field other modules need: fragile, and it has to be redone
  for every field anybody references.
- Hide the pages without removing them (mark them invisible per user): the pages stay in the
  architecture, so the client keeps reading the fields inside them and their values reach the
  agent's browser. They are removed instead.

## Note

The third-party scenario is reproduced by a test that creates the view at runtime, instead of
depending on a real module. This keeps the check self-contained and reproducible in CI.

@BinhexTeam T21484

### DIFF
--- a/sale_commission_agent_restrict/models/res_partner.py
+++ b/sale_commission_agent_restrict/models/res_partner.py
@@ -96,3 +96,16 @@ class ResPartner(models.Model):
                         res.pop("agent_ids")
                         return res
         return res
+
+    @api.model
+    def _get_view(self, view_id=None, view_type="form", **options):
+        arch, view = super()._get_view(view_id=view_id, view_type=view_type, **options)
+
+        if view_type == "form" and self.env.user.has_group(
+            "sale_commission_agent_restrict.group_agent_own_commissions"
+        ):
+            node_page_sale_purchases = arch.xpath("//page[@name='sales_purchases']")[0]
+            node_page_sale_purchases.set("invisible", "1")
+            node_page_internal_notes = arch.xpath("//page[@name='internal_notes']")[0]
+            node_page_internal_notes.set("invisible", "1")
+        return arch, view
--- a/sale_commission_agent_restrict/views/res_partner_views.xml
+++ b/sale_commission_agent_restrict/views/res_partner_views.xml
@@ -16,38 +16,4 @@
         </field>
     </record>
 
-    <record id="view_partner_form" model="ir.ui.view">
-        <field name="name">res.partner.form.agent.inherit</field>
-        <field name="model">res.partner</field>
-        <field name="inherit_id" ref="base.view_partner_form" />
-        <field name="arch" type="xml">
-            <!-- Hide these notebook pages from agents -->
-            <page name='sales_purchases' position="attributes">
-                <attribute
-                    name="groups"
-                >!sale_commission_agent_restrict.group_agent_own_commissions</attribute>
-            </page>
-            <page name='internal_notes' position="attributes">
-                <attribute
-                    name="groups"
-                >!sale_commission_agent_restrict.group_agent_own_commissions</attribute>
-            </page>
-            <!-- Add fields needed by context/domains as hidden -->
-            <page name='sales_purchases' position="before">
-                <field
-                    name="user_id"
-                    groups="sale_commission_agent_restrict.group_agent_own_commissions"
-                    invisible="True"
-                />
-                <field
-                    name="team_id"
-                    groups="sale_commission_agent_restrict.group_agent_own_commissions"
-                    invisible="True"
-                />
-            </page>
-        </field>
-    </record>
-
-
-
 </odoo>


AungKoKoLin1997:
@rrebollo I checked the root cause and the proposed fix. I don't think the change is necessary: the inherit mechanism is applied before the post-processing that removes groups= nodes, so inheriting views can safely use those pages as selectors — confirmed by the author of the upstream PR here: https://github.com/odoo/odoo/pull/95729#issuecomment-1231679462

Could you share the error log you're hitting? That would help identify the actual cause.

rrebollo:
Thanks for your review, @AungKoKoLin1997. I'll try to include a test that reveals the issue.

I think I understand your point, but I'm not talking about explicitly inheriting views. The case I'm making is for third-party addons that know nothing about this one but inherit/set a domain/some invisble condition ... using some QWeb elements under the sales/purchases pages. Such an addon might only depend on `sale`, never on `sale_commission_agent_restrict`.

As a matter of fact, if I add `sale_commission_agent_restrict` to the explicit `depends` in the manifest of the third-party addon, everything works—but that's precisely the problem. I think you can see the issue with that approach.

Nevertheless, I'll evaluate your arguments and perhaps bring a test to demonstrate the issue I'm describing.

rrebollo:
Thanks @AungKoKoLin1997 — you were right, and I owe you a correction.

Your point was that inherited views are applied **before** the group restriction, and
that the root cause described in the PR was wrong. You were correct: the original
description blamed the wrong thing. What actually breaks is the `groups` attribute
on the page.

In v16, `groups` on a node removes the node **and every field inside it** from the
architecture. Here the `groups` sat on a whole notebook page. Another module added a
field inside that page and used it in the `attrs` of a node outside the page. Because
the page's `groups` makes that field group-restricted where it is declared, while the
`attrs` makes it mandatory for everyone, Odoo rejects the view when it is validated:

```
Field '<name>' used in attrs ({'invisible': [('<name>', '=', False)]}) is restricted to the group(s) <groups>
```

That is why the module update aborted. After keeping the pages and hiding them per
user instead, the same update completes.

I also added a regression test that reproduces exactly this: it creates the offending
view at runtime (the way a user would from Studio) and asserts it can be created
without a `ValidationError`. Two more tests cover the per-user visibility and the
forms that do not contain those pages.

Thanks again for the careful review — it moved the fix in the right direction.

rrebollo:
One more piece of evidence for what this PR fixes, reproduced by hand on the **runboat of
the stable 16.0 branch** (unpatched) — the runboat hostname even embeds the branch commit,
`9e3083dc`. No Studio involved, just an inherited form view:

- a field placed inside the `Sales & Purchases` page,
- used in the `attrs` of a node outside that page.

Saving it raises exactly the diagnosed error:

> Field 'tz' used in attrs ({'invisible': [('tz', '=', False)]}) is restricted to the
> group(s) `sale_commission_agent_restrict.group_agent_own_commissions`.

Validation Error on the 16.0 runboat

The user is the default **Mitchell Admin**: no particular user is involved, because the
validation runs on the combined architecture and fails for anyone. The
`sale_commission_agent_restrict.view_partner_form` record visible in the list behind is
precisely the one this PR removes. With the patch applied, the same view saves cleanly —
which is what the new regression test asserts.

sale_commission_agent_restrict/models/res_partner.py
```diff
@@ -96,3 +103,28 @@ def _update_fields_values(self, fields):
...
+        if view_type == "form" and self.env.user.has_group(group):
+            arch = etree.XML(result["arch"])
+            nodes = arch.xpath(
+                "//page[@name='sales_purchases'] | //page[@name='internal_notes']"
+            )
+            for node in nodes:
+                node.set("invisible", "1")
+                modifiers = {}
+                transfer_node_to_modifiers(node, modifiers)
+                transfer_modifiers_to_node(modifiers, node)
```
AungKoKoLin1997: I think removing the node rather than hiding it gives the same fix with no trade-off:

- invisible keeps the fields in the arch, so the client still reads them and the values reach the agent's browser — only the rendering is suppressed. Removing them means they're never sent.
- _postprocess_access_rights removes groups nodes at this exact point in get_view, so this produces the arch agents already get on 16.0 today — no behaviour change for anyone, just the crash gone.
- The validation error is fixed either way, since it comes from groups being in the stored arch and get_view runs after validation.
- Tests would only need the assertions flipped: page absent for the agent, present for other users.

rrebollo: Thank you. I'll consider your suggestions. In the meantime, I'll move the PR back to draft until we reach a stable design. I'll ask for your review once it's "finished".

carlosdauden:
Agree with @AungKoKoLin1997 — removing the node is strictly better than hiding it here: `invisible` still ships the field values to the client, which defeats the point for a module whose job is precisely to keep agents from seeing that data. And since `_postprocess_access_rights` already does the same kind of removal for `groups=`-attributed nodes at this same point in `get_view()`, mirroring that logic manually reproduces today's exact behavior with none of the validation-time breakage.

One thing worth considering for whichever shape the mutation ends up taking: `get_view()` isn't cached, so the arch gets rebuilt on every request for every agent. `_get_view_cache_key()` is meant to be overridden for exactly this ("needing additional keys") — folding in group membership there and doing the removal inside `_get_view_cache()` instead would let the cache hold one variant per group instead of redoing that work per request:

```python
@api.model
def _get_view_cache_key(self, view_id=None, view_type="form", **options):
    key = super()._get_view_cache_key(view_id, view_type, **options)
    return key + (self.env.user.has_group("sale_commission_agent_restrict.group_agent_own_commissions"),)
```

Not a blocker, just avoids repeating the lxml work on a fairly hot path (every contact form open). Happy to look again once the design settles.

rrebollo:
Thanks @AungKoKoLin1997 and @carlosdauden — I took both suggestions. The pages are now removed from the agent's form instead of hidden, which is what the module did before, so for agents nothing changes beyond the update no longer failing. And the result is now remembered per user instead of being rebuilt on every contact form open, so we don't pay for it on a hot path. Two things worth knowing: I brought back the two hidden fields other modules read from that page (they exist for exactly this), and the list of fields sent to the agent's browser is now shorter, because the ones living on those pages are no longer part of the form — which is the point of removing them. The original failure has a regression test, and there's a second one that fails if the per-user caching is ever dropped.

carlosdauden (approved):
Ready to review and merge

rrebollo:
@carlosdauden could you remove the stale label?

carlosdauden:
I'm sorry, I'm not a PSC for the repository, so I don't have permission to do that.

AungKoKoLin1997 (approved):
Code Review: LG 👍

rrebollo:
@pedrobaeza, could you merge this based on the existing reviews? Or at least remove the "stale" label?

rrebollo:
@AungKoKoLin1997 @carlosdauden, do you consider this work worth forward-porting? I'd say yes. I'd be happy to do the work and then ask for your reviews. Would you be open to that?

carlosdauden:
I’d be happy to review it if you migrate it to version 18.0 or later

PR #679, opened 2026-06-15
[19.0][MIG] commission_formula_oca: Migration to 19.0

suker:
This PR migrates the `commission_formula_oca` module to Odoo version 19.0.

Changes:

- Bump version to 19.0.1.0.0


BhaveshHeliconia (approved):
Functional review LGTM!

BhaveshHeliconia:
@suker could you please update the title to: [19.0][MIG] commission_formula_oca: Migration to 19.0

rrebollo (approved):
Code Review. LGTM!

I've never seen the t-translation tag before. Interesting.

Hey, would you be so kind as to review my #676 in return?

commission_formula_oca/__manifest__.py
```diff
@@ -4,13 +4,13 @@
...
     "name": "Commission Formula OCA",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
     "category": "Commission",
     "license": "AGPL-3",
     "summary": "Commissions computed by formulas",
     "author": "Abstract,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/commission",
-    "depends": ["commission_oca", "account"],
+    "depends": ["account_commission_oca"],
```
pedrobaeza, line 13: Why are you changing the dependencies? Even more: in one moment, the dependencies were expanded to `account`, but the idea is that this module don't need `account` in any moment, just `commission_oca`.

pedrobaeza (changes requested):

commission_formula_oca/__manifest__.py
suker: You're right, deps shouldn't have been changed.

I could also delete `account` from deps but that would affect to tests, is ok with it?

pedrobaeza:
I think the tests in this case should only check current methods, not a flow involving invoices.

